Issue #129, opened 2016-09-08
Default to port 8080


a8bot:
<a href="https://github.com/ijsnellf"><img src="https://avatars.githubusercontent.com/u/19473346?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ijsnellf](https://github.com/ijsnellf)**
_Tuesday Sep 06, 2016 at 16:38 GMT_
_Originally opened as https://github.com/amalgam8/controller/issues/63_

---

The controller should default to port 8080 rather than 6379.

rshriram:
resolved by commit d24daf3

PR #238, opened 2016-09-12
Redis for registry

ijsnellf:
Updates the examples to use a Redis backed registry.

Also fixes several more issues with https://github.com/amalgam8/amalgam8/issues/129.


zcahana:
Was the Bluemix example intentionally left without Redis support? Looks like the controller doesn't use it as well.

ijsnellf:
@zcahana: Redis support for the Bluemix sample needs to be added. It will require more effort, so I figured it could be its own PR.

PR #245, opened 2016-09-15
Set instance ID to include <appid> for uniqueness (resolves #233)

### LINKED ISSUE #233: Eureka API: Query for a specific instance by ID fails because <appid> is missing
The query for a specific instance by ID endpoint (GET /api/eureka/{token}/v2/instances/instanceID) fails because there is no <appid> parameter in the URL path.

zachidan:
This PR reverts #235 and resolves #233


registry/api/protocol/eureka/instance_handler.go
```diff
@@ -341,13 +356,86 @@ func (routes *Routes) getInstance(w rest.ResponseWriter, r *rest.Request) {
...
+	}
+	uid := buildUniqueInstanceID("", iid)
+
+	catalog := routes.catalog(w, r)
+	if catalog == nil {
+		routes.logger.WithFields(log.Fields{
+			"namespace": r.Env[env.Namespace],
+			"error":     "catalog is nil",
+		}).Errorf("Failed to query instance %s", uid)
+
```
zcahana, line 388: Need a corresponding `i18n.Error()` too. There's a "namespace not found"-style error string defined.

registry/api/protocol/eureka/instance_handler.go
```diff
@@ -341,13 +356,86 @@ func (routes *Routes) getInstance(w rest.ResponseWriter, r *rest.Request) {
...
+			"error":     "services list is nil",
+		}).Warnf("Failed to query instance %s", uid)
+
+		i18n.Error(r, w, http.StatusNotFound, i18n.ErrorInstanceNotFound)
+		return
+	}
+
+	for _, svc := range services {
+		insts, err := catalog.List(svc.ServiceName, nil)
+		if err != nil {
```
zcahana, line 407: `catalog.List(svc.ServiceName, nil)` will return an error if the service "doesn't exist" (no instances). This can happen if the last instance has been concurrently deleted/expired. In this case, we should `continue` to search, and not return an error.

zcahana (changes requested):
See code comments

zcahana (approved):
LGTM

PR #247, opened 2016-09-15
Add eureka remote catalog

zachidan:


registry/api/protocol/eureka/application.go
```diff
@@ -14,8 +14,52 @@
...
+	"fmt"
+)
+
 // Application is an array of instances
 type Application struct {
 	Name      string      `json:"name,omitempty"`
 	Instances []*Instance `json:"instance,omitempty"`
 }
+
+// This struct is used ONLY for the specific UnmarshalJSON
```
elevran: If used only inside the UnmarshalJSON() function, you can define the types inside that function as well.

registry/api/protocol/eureka/applications.go
```diff
@@ -14,7 +14,58 @@
...
+
 // Applications is an array of application objects
 type Applications struct {
-	Application []*Application `json:"application,omitempty"`
+	VersionDelta int64
+	Hashcode     string
+	Application  []*Application
+}
+
+// This struct is used ONLY for the specific UnmarshalJSON
```
elevran: ditto - define inside the function

registry/api/protocol/eureka/applications.go
```diff
@@ -14,7 +14,58 @@
...
+type singleApplications struct {
+	VersionDelta int64        `json:"versions__delta,omitempty"`
+	Hashcode     string       `json:"apps__hashcode,omitempty"`
+	Application  *Application `json:"application,omitempty"`
+}
+
+// This struct is used ONLY for the specific UnmarshalJSON
+type multiApplications struct {
+	VersionDelta int64          `json:"versions__delta,omitempty"`
```
elevran: Consider using an embedded struct to capture VersionDelta and Hashcode. See https://play.golang.org/p/A3Fv9WW9A5

registry/api/protocol/eureka/instance_handler.go
```diff
@@ -106,60 +106,28 @@ func (routes *Routes) registerInstance(w rest.ResponseWriter, r *rest.Request) {
...
-	catalog := routes.catalog(w, r)
-	if catalog == nil {
+	si, err := BuildA8Instance(inst)
+	if err != nil {
 		routes.logger.WithFields(log.Fields{
 			"namespace": r.Env[env.Namespace],
-			"error":     "catalog is nil",
-		}).Errorf("Failed to register instance %+v", inst)
+			"error":     err,
+		}).Warnf("Failed to register instance %+v", inst)
```
elevran: why changing to warning instead of error? Because it's a user attributed to the user input?

registry/api/protocol/eureka/instance_handler.go
```diff
@@ -232,7 +200,7 @@ func (routes *Routes) deregisterInstance(w rest.ResponseWriter, r *rest.Request)
 			"error":     "catalog is nil",
 		}).Errorf("Failed to deregister instance %s", uid)
 
-		i18n.Error(r, w, http.StatusBadRequest, i18n.ErrorNamespaceNotFound)
+		i18n.Error(r, w, http.StatusInternalServerError, i18n.ErrorNamespaceNotFound)
```
elevran: If namespace not found, I would assume a bad request (unless you want to reserve that to malformed requests...).
Don't think it is a server error, maybe want to return 404?
This may be applicable in other error returns elsewhere in this change set.

registry/api/protocol/eureka/instance_handler.go
```diff
@@ -545,6 +513,47 @@ func buildExtensionFromInstance(inst *Instance) (map[string]interface{}, error)
 	return map[string]interface{}{extEureka: string(ext), extVIP: copyInst.VIPAddr}, nil
 }
 
+func BuildA8Instance(inst *Instance) (*store.ServiceInstance, error) {
```
elevran: Convert/Translate?

registry/store/eureka/client.go
```diff
@@ -0,0 +1,128 @@
...
+	eurekaapi "github.com/amalgam8/amalgam8/registry/api/protocol/eureka"
+	"github.com/amalgam8/amalgam8/registry/utils/logging"
+)
+
+const (
+	connectTimeout = time.Duration(2) * time.Second
+	requestTimeout = time.Duration(30) * time.Second
+)
+
+type eurekaClient struct {
```
elevran, line 38: Have we considered using one of the open source Eureka clients?
Not that this is a huge file so it may be ok to avoid another external dependency.

registry/store/eureka/catalog.go
```diff
@@ -0,0 +1,328 @@
...
+
+type eurekaCatalog struct {
+	client *eurekaClient
+
+	services     serviceMap
+	instances    instanceMap
+	versionDelta int64
+
+	logger *log.Entry
+	sync.RWMutex
```
elevran: would suggest putting embedded struct types at top of struct

registry/store/eureka/catalog.go
```diff
@@ -0,0 +1,328 @@
...
+		return nil, nil
+	}
+
+	// If we have the latest version, no need to do anything
+	if apps.VersionDelta == ec.versionDelta {
+		ec.logger.Infof("Delta update was skipped, because we have the latest version (%d)", apps.VersionDelta)
+		return services, instances
+	}
+
+	var updCount, delCount int
```
elevran: var updated, deleted int?

registry/store/eureka/catalog.go
```diff
@@ -0,0 +1,328 @@
...
+	var keys []string
+	for status := range hashMap {
+		keys = append(keys, status)
+	}
+	sort.Strings(keys)
+
+	for _, status := range keys {
+		count := hashMap[status]
+		hashcode = hashcode + fmt.Sprintf("%s%s%d%s", status, hashcodeDelimiter, count, hashcodeDelimiter)
+
```
elevran: empty line?

registry/store/eureka/catalog.go
```diff
@@ -0,0 +1,328 @@
...
+				insts := services[si.ServiceName]
+				if insts == nil {
+					insts = instanceMap{}
+					services[si.ServiceName] = insts
+				}
+				insts[si.ID] = si
+				updCount++
+			default:
+				ec.logger.Warnf("Unknown ActionType %s for instance %+v", inst.ActionType, si)
+
```
elevran: empty line?

registry/store/eureka/client.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		if _, err := url.Parse(u); err != nil {
+			return nil, err
+		}
+
+		urls[i] = u
+	}
+
+	hc := &http.Client{
```
elevran, line 67: Is there a need to support HTTPS and configure TLS?

elevran (changes requested):
Comments posted. Need to resolve merge conflict.

registry/README.md
```diff
@@ -118,13 +118,14 @@ Amalgam8 Registry supports a Redis backend for storing instance information as a
...
 The Amalgam8 Registry supports read-only catalogs extensions. 
-The content of each catalog extension (e.g., Kubernetes, Docker-Swarm, FileSystem, etc) is read by the Registry and
+The content of each catalog extension (e.g., Kubernetes, Docker-Swarm, Eureka, FileSystem, etc) is read by the Registry and
 returned to the user along with the content of the Registry itself.
 
 | Environment Variable | Flag Name                   | Description | Default Value |
 |:---------------------|:----------------------------|:------------|:--------------|
 | `A8_K8S_URL` | `--k8s_url` | Enable kubernetes catalog and specify the API server | (none) |
 | `A8_K8S_TOKEN` | `--k8s_token` | Kubernetes API token | (none) |
+| `A8_EUREKA_URL` | `--eureka_url` | Enable eureka catalog and specify the API server(s) | (none) |
```
elevran: if multiple servers allowed, consider renaming to eureka_urls and indicate comma separated?

registry/api/protocol/eureka/instance_handler.go
zachidan: changed to TranslateToA8Instance

registry/store/eureka/client.go
zachidan: Yes, but didn't find any good client.

In addition, we don't need the full capabilities here and we may need some watson-specific functionality (e.g., query multiple VIPs)

registry/store/eureka/client.go
zachidan: HTTPS was added

registry/README.md
zachidan: It is a multi flags, so the user can specify multiple API servers using multiple --eureka_url flags. I've updated the README.

registry/api/protocol/eureka/instance_handler.go
zachidan: If a namespace is NOT defined, we creates a new namespace. An error here means some unexpected error occurred. Thus, it is an InternalServerError.

elevran (approved):

elevran (approved):

Issue #229, opened 2016-09-08
Need Go based bookinfo microservices to speedup testing


rshriram:
The integration test is currently using the actual demo app that takes a lot of time of build (with Flask, bootstrap, etc.). Given that the integration test is running in non-interactive mode, we need Go based implementations of all the bookinfo microservices so that the app containers can be built and tested quickly. The go based test apps would also allow us to easily add additional test cases for testing A8 capabilities.

zcahana:
I'm gonna take a crack at that.

Plan is to implement an identical set of services in Golang, and delete the Python ones.
Users who'd like to modify/rebuild the examples will need a Go-enabled environment. Others will just use the prebuilt Dockerhub images.

rshriram:
You don't need identical set of services, in the sense that you don't need a bootstrap based UI, etc. The Go based version is intended only for testing. So, we could simply return JSONs/one line text from each microservice; the productpage service could simply collate the JSONs/text entries. Hopefully we would be able to use Gingo or something on the end to end system.

PR #248, opened 2016-09-15
Integration tests using Golang-based Bookinfo apps

zcahana:
- Implemented the 4 bookinfo microservices in Golang
  - Same functionality
  - JSON instead of HTML
- Modified integration test scripts to deploy the Golang bookinfo apps
- Added the integration tests to Travis build sequence. They will be run _only_ when building the `master` branch (including PRs targeting `master`).


zcahana:
Btw, this PR resolves #229

zcahana:
The integration test duration hasn't been reduced significantly. It still takes ~10 minutes (compared to ~13 minutes with Python). Time breakdown is roughly:
- 1.5 minutes compiling+building docker images (controlplane + apps)
- 1 minute running the test workload on docker
- 1 minute running the test workload on kubernetes
- Rest of the time is mostly for downloading images (particularly, ELK), starting kubernetes, and starting all containers.

Question is:
- Do we really need to re-run the same test on k8s? We can cut out several minutes without it.
- Anything else we can do to improve?

rshriram (changes requested):
@zcahana Thanks for this PR.
I have a bunch of comments.

I would like to get to a state where we easily add tests for new features we are adding to A8 (e.g., controller rules, registry catalogs, etc.), and test in integrated manner. Not sure if these can be accomplished without docker, i.e. with Gingo and friends. If so, then we can have such an integration test run for all PRs while the docker based test run for PRs from a8 folks.

Secondly, I don't think we should be depending on a8ctl while doing our regular testing. Because a8ctl is in a separate repo. We won't know if the problem is with a8ctl or with amalgam8. When a8ctl moves into amalgam8 repo, then, it makes sense to use a8ctl.
1. The test suite is only for us. So, we don't really have to follow all the microservice principles when designing the app. 

You have split the app into multiple folders and you are creating several docker containers. My suggestion is to have one bookinfo binary, where we can supply two or 3 arguments: port to listen on, serviceName and sidecar URL.
1. We build the above image out of openresty:alpine and use the same for K8S and Docker (in k8s, we just run it with a8sidecar --supervise). Takes lesser time to build
2. When running the full suite of integration tests in a frequent fashion (not the final merge with master), we can stick to only docker. And the setup can be fully trimmed to the following:
   - redis backend
   - controller and registry
   - bookinfo demo containers with a modified filebeat that logs directly to elasticsearch
   - Use rest APIs to controller to set the rules (which we can obtain from a8ctl --debug). This is where I was hoping Ginko can do the BDD tests. If not, then good old bash it is!
     - We might have to skip gremlin verification since its part of a8ctl. So, we stick with REST API based setting of fault rules and verifying that things took <7s but >5s to respond.

I realize that this is a lot of changes. We can move on its iteratively. At minimum, I suggest we do (1) [collating app into one binary], and (2) using one single docker image and running with different args for different cases, and as you noted, we can stick with Docker alone for most part.

ijsnellf (approved):
LGTM.

ijsnellf:
@rshriram: I think the changes you are recommending might warrant a new issue. In the meantime, @zcahana's changes are a vast improvement over our current automated tests.

rshriram:
@ijsnellf I agree. Lets go through with this first.

rshriram (approved):
Lets implement the comments I gave iteratively. This is good for starters. We can merge once the travis issues are resolved.

zcahana:
- Rebased my branch to eliminate merge conflicts
- Removed the Gremlin phase / ELK stack / Weaveworks stuff

PR #253, opened 2016-09-15
update swagger docs: no retries or failover

rshriram:
removed retries and failover from controller.json and schema.json. Also updated controller swagger API version to 1.0.0


ijsnellf (approved):
LGTM.

PR #254, opened 2016-09-15
Ruby details service

ijsnellf:
Replaces the implementation of the bookinfo details service with a Ruby implementation.


ijsnellf:
Currently fails in k8s. The details server port is inaccessible from outside the container.

ijsnellf:
Fixed the k8s issue by updating from `ruby:2.1` to `ruby:2.3` base images.

rshriram (approved):
LGTM. Neat work!

PR #265, opened 2016-09-19
Redis v2 changes

craigyam:
Switch to key/value strings to use automatic expiration.  Update the key to include keywords.


registry/store/external_catalog_test.go
```diff
@@ -428,55 +427,6 @@ func TestExternalListServices(t *testing.T) {
 	}
 }
 
-func TestExternalOutOfServiceDoesNotExpire(t *testing.T) {
```
zcahana, line 432: Most of the deleted test cases are still valid. I guess they were removed because the redis mock doesn't support expirations. How complex would it be to add that support?

registry/store/redis_registry.go
```diff
@@ -42,50 +42,56 @@ func NewRedisRegistry(db database.Database) ExternalRegistry {
...
+	entry, err := rr.db.ReadEntry(siKey)
 
 	if err != nil {
 		return nil, err
 	}
 	// If a key is not found, return an empty instance
-	if len(keys) == 0 {
+	if len(entry) == 0 {
 		var emptyServiceInstance ServiceInstance
 		return &emptyServiceInstance, nil
```
zcahana: Is it possible that `err == nil` but `len(entry) == 0`?
Either way, we should return an error in this case, and not an `emptyServiceInstance`.

registry/store/external_registry.go
```diff
@@ -24,18 +24,18 @@ import (
 type ExternalRegistry interface {
 	ReadKeys(namespace auth.Namespace) ([]string, error)
 	ReadServiceInstanceByInstID(namespace auth.Namespace, instanceID string) (*ServiceInstance, error)
-	ListServiceInstancesByKey(namespace auth.Namespace, key string) (map[string]*ServiceInstance, error)
+	ListServiceInstancesByName(namespace auth.Namespace, name string) (map[string]*ServiceInstance, error)
 	ListAllServiceInstances(namespace auth.Namespace) (map[string]ServiceInstanceMap, error)
 	InsertServiceInstance(namespace auth.Namespace, instance *ServiceInstance) error
-	DeleteServiceInstance(namespace auth.Namespace, key string) (int, error)
+	DeleteServiceInstance(key string) (int, error)
```
zcahana: Should the `DeleteServiceInstance()` method accept a namespace param, as all other methods in this interface, instead of expecting the client (`externalCatalog`) to encode it?
Also, what's the significance of the `int` return value? Seems to be the numeric return value of redis... which doesn't make sense as part of this interface.

registry/store/redis_registry.go
```diff
@@ -42,50 +42,56 @@ func NewRedisRegistry(db database.Database) ExternalRegistry {
...
+	err = json.Unmarshal(entry, &si)
+	if err != nil {
+		// Log an error, but continue with empty SI
+		rr.logger.WithFields(log.Fields{
+			"key": siKey,
+		}).Error("Unable to unmarshal json")
 	}
 
-	return si, nil
+	return si.DeepClone(), nil
```
zcahana: No reason to `DeepClone()` here. We've been using it for defensive copying when returning stored service instances, but in here, `si` is just a local variable that otherwise would be garbage collected.

registry/store/redis_registry.go
```diff
@@ -42,50 +42,56 @@ func NewRedisRegistry(db database.Database) ExternalRegistry {
...
 
 	for key, siBytes := range matches {
 		var si ServiceInstance
-		err := json.Unmarshal(siBytes, &si)
+		err := json.Unmarshal([]byte(siBytes), &si)
 		if err != nil {
 			// Log an error, but continue with empty SI
 			rr.logger.WithFields(log.Fields{
 				"key": key,
 			}).Error("Unable to unmarshal json")
```
zcahana, line 90: Should `continue` in case of an unmarshaling error.

registry/store/redis_registry.go
```diff
@@ -112,11 +118,11 @@ func (rr *redisRegistry) ListAllServiceInstances(namespace auth.Namespace) (map[
...
 		siMap := make(ServiceInstanceMap)
 		siKey := DBKey{
-			InstanceID:  strings.SplitN(key, ".", 2)[0],
-			ServiceName: strings.SplitN(key, ".", 2)[1],
+			InstanceID: strings.SplitN(key, keySeparator, 2)[0],
+			Namespace:  strings.SplitN(key, keySeparator, 2)[1],
 		}
 		siMap[siKey] = &si
-		serviceMap[siKey.ServiceName] = siMap
+		serviceMap[si.ServiceName] = siMap
```
zcahana: Seems to be a bug here: `serviceMap` might already contain a `ServiceInstanceMap` entry for that key (service name). In that case, it will be overwritten. The end result is that for each service, only one instance will be returned.

registry/store/redis_registry.go
```diff
@@ -125,11 +131,18 @@ func (rr *redisRegistry) ListAllServiceInstances(namespace auth.Namespace) (map[
 func (rr *redisRegistry) InsertServiceInstance(namespace auth.Namespace, instance *ServiceInstance) error {
 	// Write the JSON registration data to the database
 	instanceJSON, _ := json.Marshal(instance)
-	siKey := fmt.Sprintf("%s.%s", instance.ID, instance.ServiceName)
+	siKey := fmt.Sprintf("%s%s%s", instance.ID, keySeparator, namespace.String())
+
+	err := rr.db.InsertEntry(siKey, instanceJSON)
```
zcahana: In case of an error in `InsertEntry()`, we shouldn't `Expire()`.

registry/store/redis_registry.go
```diff
@@ -125,11 +131,18 @@ func (rr *redisRegistry) ListAllServiceInstances(namespace auth.Namespace) (map[
...
+
+	// If the status is OUT_OF_SERVICE do not expire
+	if instance.Status != OutOfService {
+		rr.db.Expire(siKey, instance.TTL)
```
zcahana: `Expire()` might fail as well. By minimum, this error should be logged.

registry/utils/database/redis_database.go
```diff
@@ -125,80 +130,88 @@ func (rdb *redisDB) ReadAllEntries(hashname string) (map[string]string, error) {
 		defer conn.Close()
 	}
 
-	entries, err := redis.StringMap(conn.Do("HGETALL", hashname))
+	entries := make(map[string]string)
+
+	matches, err := rdb.scan(conn, match)
+
+	for _, value := range matches {
```
zcahana: Calling `GET` for each instance is extremely inefficient. [`MGET`](http://redis.io/commands/mget) is a much better option.

zcahana (changes requested):
@craigyam Thanks for submitting the PR. Please see attached comments.

registry/store/external_registry.go
zcahana: Sorry, forgot to review the 2nd commit, which already corrected the namespace parameter.

registry/store/external_registry.go
```diff
@@ -27,15 +27,24 @@ type ExternalRegistry interface {
...
 type DBKey struct {
 	Namespace  string
 	InstanceID string
 }
 
+// Database key format:
+// reg:inst:<instance id>:ns:<namespace>
 func (dbk *DBKey) String() string {
-	return fmt.Sprintf("%s%s%s", dbk.InstanceID, keySeparator, dbk.Namespace)
+	return fmt.Sprintf("%s%s%s%s%s%s%s%s%s", registryRecordKey, keySeparator, instanceKey, keySeparator, dbk.InstanceID, keySeparator, namespaceKey, keySeparator, dbk.Namespace)
```
zcahana: Alternative implementation that avoid the error-prone `%s%s%s%s%s%s%s%s%s`:

``` go
parts := []string{registryRecordKey, instanceKey, dbk.InstanceID, namespaceKey, dbk.Namespace}
return strings.Join(parts, keySeparator)
```

zcahana: Also, given that there's no good reason otherwise, I would suggest to order the key elements as follows:
`reg:ns:<namespace>:inst:<instance id>`.

This should already improve the performance of regex matching during scans, and could potentially improve even fruther in the future, in case redis will support optimized scans over sorted key space.

Also, it feels more natural and hierarchical this way.

registry/store/redis_registry.go
craigyam: len(entry) == 0 is a valid response (registering a new instance for example).  I returned an empty instance because we check the instance.ID during registration.  I can instead return nil and check instance != nil.

craigyam: Done

registry/store/redis_registry.go
craigyam: Done

registry/store/redis_registry.go
craigyam: Done

registry/store/redis_registry.go
craigyam: Done

registry/store/redis_registry.go
craigyam: Done

registry/store/redis_registry.go
craigyam: Done

registry/store/external_registry.go
craigyam: The return value is the number of entries deleted (0 is a valid value).  The calling code checks if the delete returned 0 and will respond that there was no service to delete.

registry/utils/database/redis_database.go
craigyam: Done

registry/store/external_registry.go
craigyam: So I changed the DB key to be a JSON string which makes parsing the key read from the db way easier (just an Unmarshal vs. splitting the string based on the separators - the concern being handling separators that may appear in the values).  This should make adding/removing/moving fields in the key trivial compared to maintaining a parsing function.  Let me know what you think of the change.  If you prefer it the other way, I can change it back.

craigyam: See above comment.

zcahana: Assuming we'll have to support existing keys - adding/removing/moving fields in the key will be impossible, and at minimum not trivial at all. That also means that we must have control over the JSON marshaling - whitespace, fields ordering. `json.Marshal()` doesn't specify that, so we'll have to implement `MarshalJSON()` ourselves, including escaping `"` characters from the values etc. That also means larger keys (JSON overhead) and extra work to encode. The only benefit is we won't be needing to implement decoding on our own - the off-the-shelf `json.Unmarshal()` will do fine - but given the other issues, I suggest we stick to a simple (non-JSON) encoding scheme, and escape the `:` separators if present.

zcahana: On another note, need to take into account that when doing SCAN, the namespace might contain glob-characters as well (`*`, `?`, ...). These must be escaped as well.

craigyam: Switched back to the original key format and added a parser.  Also added a function to escape the glob chars.

craigyam: Done

zcahana (approved):

zcahana:
@craigyam Looks good, thanks!

Issue #266, opened 2016-09-19
Sidecar fails to update Nginx with Go 1.7.1


zcahana:
Following the upgrade to Go 1.7.1 (#263), I'm seeing the following errors in the gateway's logs when running the integration tests:

```
time="2016-09-19T09:36:16Z" level=error msg="Failed to send request to NGINX server" error="Post http://localhost:5813/a8-admin: dial tcp: i/o timeout" 
time="2016-09-19T09:36:16Z" level=error msg="Failed to update HTTP upstreams with NGINX" error="Post http://localhost:5813/a8-admin: dial tcp: i/o timeout" 
time="2016-09-19T09:36:16Z" level=warning msg="Controller listener failed" error="Post http://localhost:5813/a8-admin: dial tcp: i/o timeout" 
time="2016-09-19T09:36:46Z" level=error msg="Failed to send request to NGINX server" error="Post http://localhost:5813/a8-admin: dial tcp: i/o timeout" 
time="2016-09-19T09:36:46Z" level=error msg="Failed to update HTTP upstreams with NGINX" error="Post http://localhost:5813/a8-admin: dial tcp: i/o timeout" 
time="2016-09-19T09:36:46Z" level=warning msg="Registry listener failed" error="Post http://localhost:5813/a8-admin: dial tcp: i/o timeout" 
```

... which shows that the sidecar isn't able to properly set up a TCP connection with the Lua server running within Nginx on port 5813. Manually exec'ing into the gateway container, I can successfully run `curl localhost:5813/a8-admin` repeatedly.

Note that I've tested this multiple times with Go 1.6.2, 1.7 and 1.7.1. This is repeatedly happening with Go 1.7.1, but not on 1.6.2 nor 1.7, so circumstantially I'm tracing this to Go 1.7.1.

zcahana:
Further investigation shows that this is related to DNS resolution of "localhost" (as part of `POST http://localhost:5813`) within the sidecar container. Note that this happens only when running within K8s:
- On Go 1.7.1, it just times out after 30s (hence the `dial tcp: i/o timeout` above)
- On Go 1.7, it succeeds, but DNS resolution takes 20s.
- On Go <= 1.6.3, it succeeds immediately.

This seems to be related to golang/go#16865, which supposedly fixed a regression in Go 1.7 around DNS resolution logic.

zcahana:
@ericvn Note that we've temporarily reverted back to Go 1.6.3, until the issue is resolved.

[...]

PR #267, opened 2016-09-19
Revert to Go 1.6.3

zcahana:
Revert to Go 1.6.3 (from Go 1.7.1) until issue #266 is further investigated and resolved.


rshriram (approved):

ericvn:
Looks good. 1.6.3 is supported with Sierra on Macs so is a good solution until the 1.7.1 issues can be resolved.

PR #275, opened 2016-09-20
Remove blank lines and add --anti flag to deploy-bookinfo.sh

dcberg:
The deploy to Bluemix was failing when there were empty lines in the "bluemix ic group-create" command. 
Remove the image tag from line 52 because it was already included in the variable.
Added the "--anti" flag (anti-affinity) as a best practice. This does require the latest version of the bluemix cli to be installed.


rshriram (approved):
Looks good. Thanks @dcberg

PR #278, opened 2016-09-20
split nginx into multiple config files

### LINKED ISSUE #259: Split amalgam8-services.conf into multiple files
amalgam8-services.conf has lot of A8 specific stuff as well as parts that the user can customize. To prevent accidental misconfiguration of A8 specific stuff, we need to move all A8 specific configuration into individual configuration files and keep the file simple. 

An example can be found in GameOn Proxy. https://github.com/gameontext/gameon-proxy/pull/16

rshriram:
resolves #259


zcahana (approved):
LGTM

PR #284, opened 2016-09-20
Increase length for input endpoint value on registration and output better messages for invalid inputs

dougbeauchene:


dougbeauchene:
If I  `git clone https://github.com/dougbeauchene/amalgam8.git`

Run `make verify` it errors like the build does

Then run `make precommit` it is successful

Then run `make verify` again it is successful

I have no idea what is wrong

craigyam:
Looks like there are a few tabs at the end of line 430 in instance_handler.go.

registry/api/protocol/amalgam8/instance_handler.go
```diff
@@ -72,7 +72,22 @@ func (routes *Routes) registerInstance(w rest.ResponseWriter, r *rest.Request) {
...
+			case store.ErrorInstanceEndpointValueTooLong:
+				i18n.Error(r, w, statusCodeFromError(err), i18n.ErrorEndpointValueTooLong, valueMaxLength)
+			case store.ErrorInstanceStatusLengthTooLong:
+				i18n.Error(r, w, statusCodeFromError(err), i18n.ErrorStatusLengthTooLong, statusMaxLength)
+			case store.ErrorInstanceMetaDataTooLong:
+				i18n.Error(r, w, statusCodeFromError(err), i18n.ErrorMetaDataTooLong, metadataMaxLength)
+			default:
+				i18n.Error(r, w, statusCodeFromError(err), i18n.ErrorInstanceRegistrationFailed)
+			}
+		}
```
zcahana, line 90: It would be safer to add an `else` block here, for the case that `regerr` is not of type `*store.Error`. In that case, we would want to return the generic `i18n.ErrorInstanceRegistrationFailed` error message as well.

registry/messages/en-US.json
```diff
@@ -138,5 +138,25 @@
...
     "id": "eureka_error_vip_required",
     "translation": "vip is required"
+  },  
+  {
+    "id": "error_instance_service_name_not_specified",
+    "translation": "Failed to register instance because service name value was not specified"
+  },  
+  {
+    "id": "error_instance_service_name_too_long",
+    "translation": "Failed to register the instance because service name value exceeded {{.Count}} characters"
```
zcahana, line 148: To be precise, the limit is on the number of _bytes_, not _characters_.

registry/api/protocol/eureka/instance_handler.go
```diff
@@ -168,7 +168,22 @@ func (routes *Routes) registerInstance(w rest.ResponseWriter, r *rest.Request) {
...
+			case store.ErrorInstanceEndpointValueTooLong:
+				i18n.Error(r, w, http.StatusBadRequest, i18n.ErrorEndpointValueTooLong, valueMaxLength)
+			case store.ErrorInstanceStatusLengthTooLong:
+				i18n.Error(r, w, http.StatusBadRequest, i18n.ErrorStatusLengthTooLong, statusMaxLength)
+			case store.ErrorInstanceMetaDataTooLong:
+				i18n.Error(r, w, http.StatusBadRequest, i18n.ErrorMetaDataTooLong, metadataMaxLength)
+			default:
+				i18n.Error(r, w, http.StatusBadRequest, i18n.ErrorInstanceRegistrationFailed)
+			}
+		}
```
zcahana, line 186: Same comment applies here.

zcahana (approved):
Thanks @dougbeauchene for submitting this.
There are several minor changes needed to be done (see review comments), but given that @dougbeauchene will be out till Monday and these are very minor changes, I'm gonna merge this PR as is, and submit another one with the fixes.

Issue #50, opened 2016-09-08
Deregister from registry upon app termination


a8bot:
<a href="https://github.com/zcahana"><img src="https://avatars.githubusercontent.com/u/17428696?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zcahana](https://github.com/zcahana)**
_Wednesday Aug 10, 2016 at 12:55 GMT_
_Originally opened as https://github.com/amalgam8/sidecar/issues/50_

---

When the sidecar process is about to terminate (either due to its supervised app terminating, or a SIGTERM/SIGINT), it should attempt to deregister its app from the registry, instead of just letting it expire (which happens in up to <TTL>). This could speed up upstream reconfiguration in dependent services, and avoid failed requests to this no-longer-valid app instance.

zcahana:
This has not been resolved yet (despite the comment on #294).

Issue #51, opened 2016-09-08
Add a healthcheck mechanism


a8bot:
<a href="https://github.com/zcahana"><img src="https://avatars.githubusercontent.com/u/17428696?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zcahana](https://github.com/zcahana)**
_Wednesday Aug 17, 2016 at 13:01 GMT_
_Originally opened as https://github.com/amalgam8/sidecar/issues/51_

---

Add a healthcheck mechanism that periodically polls the health of the supervised app, and stop heartbeating to the Registry while the app is unhealthy.

Healthcheck types can be:
1. HTTP (i.e., issue an HTTP GET to some endpoints and check the return code)
2. TCP (i.e., open a TCP connection to some endpoint, and verify connection is established successfully)
3. Process (i.e., launch a process [e.g., script that greps app logs], and examine its exit status)

ijsnellf:
I'm going to be on vacation, so I'm handing this over to @GregHanson.

PR #294, opened 2016-09-22
HTTP health checks in sidecar

ijsnellf:
Adds simple HTTP health checks to the sidecar registration logic as described in https://github.com/amalgam8/amalgam8/issues/51. No TCP or shell script health check implementations are provided.

Health checks are performed asynchronously before registration and renewal. The failure of any health check will result in the sidecar de-registering the instance. Health checks are bounded by the interval between heartbeats --if a health check does not return within that time, the health check is considered to have timed out and the app is considered unhealthy. This is a very basic design that will need to be improved upon.


rshriram (changes requested):
What about K8S examples? You can use the same health check endpoints there as well. Its slightly redundant to do health checks in K8S, but we can first keep the examples consistent by adding the health checks. Then we can cleanup the K8S case by removing the registration completely and falling back to the K8S adapter.

sidecar/config/flags.go
```diff
@@ -36,6 +36,7 @@ const (
 	controllerTokenFlag = "controller_token"
 	controllerPollFlag  = "controller_poll"
 	superviseFlag       = "supervise"
+	healthChecksFlag    = "health_checks"
```
zcahana: Flag name (`health_checks`) is different than the corresponding YAML key (`healthchecks`).
For consistency, these should be the same. I personally prefer `healthchecks` as it reads cleaner.

sidecar/config/config.go
```diff
@@ -66,8 +66,9 @@ type Config struct {
 	Registry   Registry   `yaml:"registry"`
 	Controller Controller `yaml:"controller"`
 
-	Supervise bool     `yaml:"supervise"`
-	App       []string `yaml:"app"`
+	Supervise    bool     `yaml:"supervise"`
+	App          []string `yaml:"app"`
+	HealthChecks []string `yaml:"healthchecks"`
```
zcahana: Need to consider if we want each healthcheck to be a simple string, or an healthcheck object.
For example:

``` yaml
healthchecks:
    - type:     http
      value:    http://localhost:9080/health
      interval: 10

    - type:     http
      value:    http://localhost:9080/health2
      method:   POST
      code:     201
      interval: 30

    - type:     command
      value:    /bin/sh -s "cat /opt/myapp/app.log | tail -n 1000 | grep -i ERROR"
      intreval: 60

    - type:     tcp
      value:    localhost:5555
      timeout:  3
      interval: 30  
```
- Not saying we should implement all these customizations, but maybe leave room for?

For flags / environment variables, we can support only a subset of the customizations, so that it won't get too cumbersome to define. We can expect / drive users to define the healthchecks via the config file, baked into their app images.

sidecar/register/registration.go
```diff
@@ -82,12 +83,14 @@ func (agent *RegistrationAgent) Stop() {
 
 func (agent *RegistrationAgent) register() {
 	for {
-		registeredInstance, err := agent.config.Client.Register(agent.config.ServiceInstance)
-		if err == nil {
-			go agent.renew(registeredInstance)
-			return
+		if agent.checkAppHealth(DefaultReregistrationDelay) {
```
zcahana: I'd like to suggest an alternative approach for combining the healthchecking / registration logic:
1. Leave the `RegistrationAgent` as is (before the PR).
2. Implement a `Healthchecker` object that continuously executes the configured healthchecks on the background.
3. The `Healthchecker` is either in healthy or unhealthy state at any given time, depending on the combined results of the healthchecks.
4. When the `Healthchecker` moves to healthy state, it immediately invokes `agent.Start()`. Similarly, when it moves to unhealthy state, it immediately invokes `agent.Stop()`.

What we gain:
1. Simpler state machine. Especially if we add more advanced stuff like different intervals for healthchecks, heartbeats, TTL, and grace period for failed healthchecks.
2. Faster response (registration/deregistration) to becoming healthy/unhealthy

zcahana (changes requested):

sidecar/register/registration.go
rshriram: +1

ijsnellf: I was also considering that design. My concern was that health checks could be as stale the health check interval, somewhere between [0, 20s) in the case where the health check interval is the same as the default heartbeat interval. However, as you point out, it would decouple the health check and heartbeat intervals, which is probably of greater value than ensuring the health check is as fresh as possible. With that consideration, I think I agree that that design would be an improvement over the existing design.

sidecar/config/config.go
ijsnellf: Agree, that's a good idea.

sidecar/register/registration.go
ijsnellf: Also had not considered starting and stopping the registration agent --that is a good idea.

ijsnellf: Ah, never mind. By starting and stopping the registration agent from the health checker, you can ensure that the health state information is transmitted as soon as it is detected. That is a vast improvement.

sidecar/config/config.go
GregHanson: Quoted from Slack in #devel:

> There are a few options for the command line arguments. I'd like to have people's input.
> 1. Only support 1 health check definition (if you wanted more than 1, you'd have to define them in the `config.yaml`):
> `a8sidecar --healthcheck_type http --healthcheck_value http://abc:5000/health --healthcheck_timeout 30s --healthcheck_interval 15s`
> 2. Support only basic health check definitions (more advanced definitions would have to be in `config.yaml`):
> `a8sidecar --healthcheck http://abc:5000/health,http://xyz:5111/health2`
> 3. Support complete health check definitions with JSON:
> `a8sidecar --healthcheck '{"type":"http", "value": "http://abc:5000", "timeout":"30s", "interval":"15s"}' --healthcheck '{"type":"command", "/bin/sh service x status"}'`
> 
> Thoughts?

ijsnellf:
As per @zcahana's recommendations, moved to a [separate HealthChecker](https://github.com/amalgam8/amalgam8/pull/294#discussion_r80382502) implementation and partially addressed the health check [config options](https://github.com/amalgam8/amalgam8/pull/294#discussion_r80382102) suggestion. However, I did not manage implement the changes to the config to support the new configuration options. For this reason, the PR is not ready to merge.

zcahana:
I added support for passing-in the configuration options discussed above (http only, currently).

Things still missing in this PR, before being ready for merge:
- Validation/defaults for the configuration
- Tests (both for configuration as well as the healthcheck implementation)
- Solve merge conflicts

Things for next PRs:
- TCP / command health checks
- Convert examples to use config files for more elaborate healthchecks

zcahana:
Implemented validation/defaults logic for the HTTP healthchecks, and added unit tests for healthchecks configuration. Still missing is testing of the healthcheck logic/implementation itself.

One option to make this more testable and is to have the `HealthCheck` be in charge of a single healthcheck only (i.e., `Execute()` instead of `Start()` and `Stop()`), and move the responsibility for continuously driving it to the `HealthChecker`. This will also make it more simple to add additional type of healthchecks later.

zcahana (approved):

zcahana:
Build is failing due to a bad commit that bypassed Travis checks (see #313).
Once the fix (#315) is merged to `master`, we can merge `master` back into this branch (or rebase this branch on top of `master`), to get a good build.

rshriram:
At some point, we need to migrate the K8S examples to stop using explicit registration and health checks completely. Otherwise, this LGTM.

rshriram (approved):

GregHanson:
resolves Issue #50

Issue #295, opened 2016-09-22
Bluemix example: Verify if the container groups were created successfully.


gicaceres:

ijsnellf:
Resolved by #296.

PR #296, opened 2016-09-22
Verify if the container groups for the Bookinfo app were created succ…

gicaceres:
…essfully in the bluemix example.

https://github.com/amalgam8/amalgam8/issues/295


ijsnellf:
What motivated the change from 256mb to 128mb on the containers?

ijsnellf:
Resolves #295.

ijsnellf:
The change to 128mb is so that the example runs under the default Bluemix quota.

ijsnellf (approved):
LGTM

Issue #308, opened 2016-09-28
Sidecar fails when A8_SERVICE env variable is not set


zachidan:
1. If the A8_SERVICE is not set, the nginx process fails to create the Amalgam8 lua object and then most of the calls to nginx fail with a nil pointer exception. A special validity check (including user-friendly error message) for the A8_SERVICE is required on startup.
2. The sidecar allows the user to provide command line arguments and/or env variables (e.g. --service or A8_SERVICE), but the nginx requires the user to provide ONLY env variables (nginx does NOT read the command line arguments of the sidecar). We need to either convert the command line arguments to env variables on the sidecar or to disable the command line arguments on the sidecar.

rshriram:
@zachidan 
(2) is a real problem. Sidecar needs to set the A8_SERVICE environment variable before forking nginx.

PR #320, opened 2016-10-05
check for nil A8 object while initializing

### LINKED ISSUE #319: No error message issued from Lua when A8_SERVICE is not set
Nginx+Lua silently fails without issuing an error message stating that the A8_SERVICE variable is not set. Subsequent requests to obtain the internal state or even route API calls via the sidecar fail with HTTP 500. It would be better if Lua can throw an error at the very least. Better still, if it can shutdown nginx altogether, it would make things easy.

rshriram:
Fixes #319


zachidan (approved):

rshriram:
Also addresses #308

PR #321, opened 2016-10-05
allow actions without a backend (partial)

### LINKED ISSUE #311: Action rule without route rule is ignored
When an action rule is defined (e.g., delay injection to request) with no route rule, the action is ignored.

I can think about scenario that the user want to do some action for each request (e.g., delay injection, request modification, response modification, etc.)

rshriram:
fixes #311


zachidan (approved):

rshriram:
To note: when an action with tags is specified and no backend is provided, the action will be ignored.

rshriram:
actions with tags are now fully supported, even if no backend is selected. Actions are now matched with instance tags as of PR #334

PR #325, opened 2016-10-06
Translates eureka metadata tags into amalgam8 tags (closes #317)

### LINKED ISSUE #317: Eureka adapter: allow eureka users to set Amalgam8 tags for version-based routing
The sidecar supports version-based routing by using instance tags.

To allow eureka users to use version-based routing, it is required to allow them to set the tags array. My suggestion is to map the **eureka.metadata.tags** instance property to the Amalgam8 **tags array** instance property.

zachidan:


registry/api/protocol/eureka/instance_handler.go
```diff
@@ -183,9 +166,18 @@ func (routes *Routes) registerInstance(w rest.ResponseWriter, r *rest.Request) {
...
-	return json.Unmarshal(jsonString, &js) == nil
+func extractTagsFromMetadata(jsonString json.RawMessage) ([]string, error) {
+	var md map[string]interface{}
+	if err := json.Unmarshal(jsonString, &md); err != nil {
+		return nil, err
+	}
+
+	var tags []string
+	if val, ok := md[metadataTags]; ok {
+		sval := fmt.Sprintf("%v", val)
```
zcahana: `sval := fmt.Sprintf("%v", val)` - not sure why we need this?
How about just downcasting to `string`? (and making sure it's ok)

registry/api/protocol/eureka/instance_handler.go
```diff
@@ -183,9 +166,18 @@ func (routes *Routes) registerInstance(w rest.ResponseWriter, r *rest.Request) {
...
+		tags = strings.Split(sval, metadataTagsDelimiter)
```
zcahana: `tags = strings.Split(sval, metadataTagsDelimiter)`
I think we need to trim whitespaces around the split tags, so that users will be able to specify them as `tag1, tag2`.

zcahana (commented):
All in all the code looks ok, but a simple unit-test to test this could be nice.
Also note that our Eureka-integration doc on amalgam8.io needs to be updated with this.

registry/api/protocol/eureka/instance_handler.go
zachidan: @zcahana The JSON format of the metadata is map[string]interface{} and NOT map[string]string. The Netflix/Eureka encodes string as JSON string and number as JSON number. Therefore, we need to convert the value to string.

registry/api/protocol/eureka/instance_handler.go
zachidan: @zcahana Good idea. btw, note that the delimiter is semi colon and not comma, because comma is reserved char in archaius configuration.

zcahana: Weird, I also wrote a comment about the delimiter, but apparently it disappeared from the review...
I was looking at [Spring Cloud](http://cloud.spring.io/spring-cloud-netflix/spring-cloud-netflix.html), and saw that `,` is normally used as a delimiter (search for "comma-separated"). However, if that clashes with Archaius, than we can just use `;`.

zcahana (approved):

PR #326, opened 2016-10-06
Added container labels

CarltonSemple:
Added an amalgam8.role label to containers that is to be used by [weave scope](https://github.com/weaveworks/scope) in the filtered view


CarltonSemple:
When running a lot of containers, it gets very cluttered when trying to view them in a tool like [weave scope](https://github.com/weaveworks/scope). The idea of this PR is that when Weave Scope implements custom filters by container labels, Amalgam8-related containers can be shown neatly by themselves

frankbu:
@CarltonSemple: I'm wondering if it would be better, at least for now, for you to just create your own Dockerfiles that extend from the Amalgam8 ones and adds the label you need. It seems like it may be premature to make changes to the A8 example images themselves, before we have a clear understanding of why and when A8 users would/should need to do the same in their images.

frankbu:
The problem with adding labels in a8 examples is that they appear to adding an extra requirement on app writers that want to integrate with a8 (assuming that the examples are showing app writers what they need to do). Unless we say that the label needs to be there and explain why, we're just confusing people with another thing that they won't know what to do with. We don't want to add something that impacts app writers without documenting what it's for. So I think doc changes would be required, at a minimum. What do you think @rshriram?

rshriram:
I agree. I made similar comments on missing documentation in the other PR. 

One alternative may be to create a separate repo that showcases integrations with A8. These integrations can include monitoring, visualizations, etc. All with documentation of course. This way, in the demo walk throughs and documentations, we can point people to various integrations during the walk through.

CarltonSemple:
@rshriram how about a comment next to the labels explaining that they're optional? Much simpler solution. The idea being that the examples should work seamlessly with Scope when the filter is enabled. If you make things too difficult to see, people won't make the effort or even know to try it.

CarltonSemple:
I closed the other pull request because I agree that the plugin should be separate. However, suppose someone is in this repository and they run the demo apps, then go to the other repository to run Scope with the plugin? I say this because I've been doing this very thing, as an end user.

Planning for a walkthrough is nice, but you really have to consider the possibility that people will try this without a walkthrough. I think our views are different because we're coming from two different ends. You as the creators, me as the end user

PR #327, opened 2016-10-06
Weavescope routing plugin

CarltonSemple:
The plugin shows the Amalgam8 routing weight for versions of services that have had the weights specified, and not default versions. It only attaches to the containers that belong to the service


rshriram:
@CarltonSemple, thanks for this PR. It would be nice to have some more context about this code. Where does it belong, purpose, documentation, etc. 

May be it's best to have an integrations repository and put this code there?

A couple of screenshots might be helpful. 
General comments on the PR:
 No .DS_Store files. No binaries. No unnecessary files in general.

rshriram (changes requested):
See my earlier comments

PR #329, opened 2016-10-10
Nginx: Support service urls without requiring a trailing slash

### LINKED ISSUE #260: Nginx location block regex must support services without trailing slash
Current regex used in amalgam8-services.conf assumes that there will be a trailing slash after the service_name in teh location block. This may not always be the case and might cause nginx to return 404 to the user.

`location ~ ^/(?<service_name>)/(?<reqpath>.*)`  only works if url is /foo/fooAPI. But it won't work if url is just /foo.

rshriram:
fixes #260


GregHanson (approved):
LGTM

PR #333, opened 2016-10-11
Fix link to demos in README

markstgodard:
When trying out the demo apps, I noticed the main link to demos on README.md is broken


rshriram (approved):
Thanks!

Issue #255, opened 2016-09-15
a8sidecar needs to provide a debug option to inspect routes in lua


rshriram:
Need to provide commands like `a8sidecar inspect-routes` that should display the list of routes, rules etc., present in Lua. For this, a8sidecar should query lua and print it out for the user. This is much more user friendly than asking the user to query http://localhost:5813/a8-admin.

GregHanson:
@rshriram are you suggesting a CLI for the sidecar, where the user will have to log into the sidecar container and run the commands?

rshriram:
Yes. The curl we do, to inspect the lua state can be replaced by the cli. 
Instead of creating a new tool, we can just have a debug command like a8sidecar --debug show-lua-state or something along those lines. 

The cli can do more in future. Inspect lua state as well as compare it with sidecar's internal state.

rshriram:
resolved by #335

PR #335, opened 2016-10-11
cli for debugging sidecar state. 

GregHanson:
shows sidecar cached data and lua cached data

example usage (docker exec -it gateway sh):
a8sidecar --debug show-state

### DIFF
--- a/sidecar/api/debug.go
+++ b/sidecar/api/debug.go
@@ -1,0 +1,66 @@
+// Copyright 2016 IBM Corporation
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+package api
+
+import (
+	"net/http"
+
+	"github.com/amalgam8/amalgam8/controller/rules"
+	"github.com/amalgam8/amalgam8/registry/client"
+	"github.com/amalgam8/amalgam8/sidecar/proxy"
+	"github.com/ant0ine/go-json-rest/rest"
+)
+
+// DebugAPI handles debugging API calls to sidecar for checking state
+type DebugAPI struct {
+	nginxProxy proxy.NGINXProxy
+}
+
+// NewDebugAPI creates struct
+func NewDebugAPI(nginxProxy proxy.NGINXProxy) *DebugAPI {
+	return &DebugAPI{
+		nginxProxy: nginxProxy,
+	}
+}
+
+// Routes for health check API
+func (d *DebugAPI) Routes(middlewares ...rest.Middleware) []*rest.Route {
+	routes := []*rest.Route{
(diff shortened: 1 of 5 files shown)


GregHanson:
proposed fix for issue #255

sidecar/api/debug.go
```diff
@@ -0,0 +1,66 @@
...
+		rest.Get("/state", d.checkState),
+	}
+
+	for _, route := range routes {
+		route.Func = rest.WrapMiddlewares(middlewares, route.Func)
+	}
+	return routes
+}
+
+// GetHealth performs health check on controller and dependencies
```
rshriram: copy pasted comment?

sidecar/config/config.go
```diff
@@ -85,6 +85,8 @@ type Config struct {
 	LogstashServer string `yaml:"logstash_server"`
 
 	LogLevel string `yaml:"log_level"`
+
+	Debug string `yaml:"debug"`
```
rshriram: does it even make sense to have the debug flag in the YAML? can we get away without defining it here?

sidecar/sidecar.go
```diff
@@ -220,5 +232,126 @@ func startProxy(conf *config.Config) error {
...
+}
+
+func cliCommand(command string) {
+
+	switch command {
+	case "show-state":
+		httpClient := http.Client{
+			Timeout: time.Second * 10,
+		}
+		req, err := http.NewRequest("GET", "http://localhost:6116/state", nil)
```
rshriram, line 292: is it possible to externalize this URL? might come in handy for remote debugging, if the need ever arises.

sidecar/sidecar.go
```diff
@@ -220,5 +232,126 @@ func startProxy(conf *config.Config) error {
...
+			Rules     []rules.Rule                     `json:"rules"`
+		}{}
+
+		err = json.Unmarshal(respBytes, &sidecarstate)
+		if err != nil {
+			fmt.Println("Error occurred loading JSON response:", err.Error())
+			return
+		}
+
+		req, err = http.NewRequest("GET", "http://localhost:5813/a8-admin", nil)
```
rshriram, line 321: same as above. May be in the YAML file under the debug_cli, we could mention these two urls (and as part of cli flags, we can mention these two urls)

rshriram (changes requested):

sidecar/config/config.go
GregHanson: yes, I agree, since this flag is only for the CLI, it would not make since to define a yaml flag for it since yaml flags are only read once on start up

sidecar/sidecar.go
GregHanson: This sidecar is only ever checking it's own internal state state - so why should the URL be external?  What is the benefit?

rshriram: externalizing the URL is always a better practice, rather than hardcoding it. If one were to change the nginx admin port from 5813 to some other port (as they have another app listening on that port), they would have to recompile the sidecar as well, just to be able to get A8 working (and the debug stuff enabled).

Secondly, it would allow us to remotely debug the sidecar and nginx, if the user decides to expose these two admin ports outside of the container. The user would have to just use the stock a8sidecar binary and point to these two urls, and they can obtain all the internal state inside these entities.

rshriram: On second thought, the externalized case looks more like a corner case, where one wants to debug the complete system from a remote terminal. So, lets ignore that. I will create a separate issue to externalize all hard coded URLs in the sidecar code.

rshriram (approved):

rshriram:
fixes #255

PR #336, opened 2016-10-12
fix curl command in sidecar installation script

rshriram:
curl command in sidecar installation script was not honoring github redirects. Fixed by adding -L flag.


craigyam (approved):

craigyam (approved):
Trying again...

zcahana (approved):
LGTM

Issue #52, opened 2016-09-08
Extend app supervision function to support additional processes


a8bot:
<a href="https://github.com/zcahana"><img src="https://avatars.githubusercontent.com/u/17428696?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zcahana](https://github.com/zcahana)**
_Wednesday Aug 17, 2016 at 13:07 GMT_
_Originally opened as https://github.com/amalgam8/sidecar/issues/52_

---

The sidecar currently supports a single configurable app to supervise, and has a hard-coded logic to launch filebeat (as well as nginx). We'd like to have a more generalized mechanism in which the sidecar can be configured with multiple processes to start: the "main" app process, and secondary "helper" processes, like log shipping agent, log rotate agent, metrics agent, etc.

rshriram:
resolved by #349

PR #340, opened 2016-10-13
Remove filebeat from sidecar installation script

rshriram:
Filebeat is used to ship nginx logs to elastic search. It serves its purpose only in the example scripts. For sidecar installation in any arbitrary application, it is incorrect to assume that ELK stack + Filebeat will be used to ship logs. So filebeat has been removed from the sidecar installation script and the examples have been updated accordingly.

The new sidecar installation script can be used in Debian-based and Centos/Fedora/RHEL distributions (not alpine).

Note that the standalone sidecar image used in K8S will still contain filebeat, until #52 is addressed.

The A8_LOG options will be removed from the documentation in amalgam8.github.io, as we do not want users to be using this option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,13 +204,11 @@ release.sidecar:
 	@mkdir -p $(RELEASEDIR) $(BUILDDIR) \
 		$(BUILDDIR)/opt/a8_lualib \
 		$(BUILDDIR)/opt/openresty_dist \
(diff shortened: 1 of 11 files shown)


frankbu (approved):

GregHanson (approved):
LGTM

PR #341, opened 2016-10-13
Pass the request headers from the auth middleware to authenticators.

craigyam:
Update the authenticators to accept a context.  The auth middleware will pass the request headers to the authenticator using the context.


pkg/auth/authenticator.go
```diff
@@ -14,12 +14,19 @@
 
 package auth
 
-// Module name to be used in logging
-const module = "AUTH"
+import "golang.org/x/net/context"
```
zcahana: We should be using the [standard library's context package](https://golang.org/pkg/context/).
It was only introduced in Go 1.7, so we'll need to upgrade our Go version for that.

Given that our previous issues with Go 1.7 has already been resolved (see #266), I'll submit a PR with the Go update, so please rebase your branch on that update.

pkg/auth/trusted.go
```diff
@@ -23,7 +25,7 @@ func NewTrustedAuthenticator() Authenticator {
 	return trustedAuth
 }
 
-func (aut *trustedAuthenticator) Authenticate(token string) (*Namespace, error) {
+func (aut *trustedAuthenticator) Authenticate(token string, ctx ...context.Context) (*Namespace, error) {
```
zcahana: By convention, the context should be the first parameter of a function (see [here](https://golang.org/pkg/context/)).
This means that the `...` vararg will also need to go away.

registry/api/middleware/auth.go
```diff
@@ -54,7 +56,10 @@ func (mw *AuthMiddleware) handler(writer rest.ResponseWriter, request *rest.Requ
 		token = parts[1]
 	}
 
-	nsPtr, err := mw.Authenticator.Authenticate(token)
+	// Add the headers to the context passed to the authenticator
+	ctx := context.WithValue(context.Background(), auth.ContextHeadersKey, request.Header)
```
zcahana: The intended pattern is NOT to initialize a new context with `context.Background()`, but rather to fetch the existing context already associated with the http request. However and unfortunately, the `go-json-rest` package doesn't support context (yet?), and neither exposes the underlying `http.Request`. My suggestion for an interim solution is to add a `ContextMiddleware`, somewhere along the top of the middlewares stack, that initializes a context and puts it in the request's `Env` map, with a well-known key. Later on, we should probably ditch `go-json-rest` in favor a better http framework.

zcahana (changes requested):

pkg/auth/authenticator.go
zcahana: PR #345 updates go Go 1.7.1. Once merged to master, please rebase.

zcahana:
P.S. I'm not sure what caused the Travis build failure, but retriggering it seems to have turned it green :)

craigyam:
Tried to add another commit to address the code review comments but it seems to have done something funny.  Will investigate and work to clean it up.

zcahana (approved):

zcahana:
LGTM, thanks @craigyam.

PR #343, opened 2016-10-13
sidecar: return HTTP 503 where appropriate

rshriram:
When no instances of a service exist, but valid routes to the service are present, the sidecar should return HTTP 503 instead of HTTP 404.

### DIFF
--- a/sidecar/nginx/lua/amalgam8.lua
+++ b/sidecar/nginx/lua/amalgam8.lua
@@ -676,11 +676,6 @@ function Amalgam8:apply_rules()
    ngx.var.a8_upstream_name = destination
 
    local instances = get_unpacked_val(ngx_shared.a8_instances, destination)
-   if not instances or #instances == 0 then
-      ngx.status = ngx.HTTP_NOT_FOUND
-      ngx.exit(ngx.status)
-   end
-
    local routes = get_unpacked_val(ngx_shared.a8_routes, destination)
    local actions = get_unpacked_val(ngx_shared.a8_actions, destination)
    local headers = ngx.req.get_headers()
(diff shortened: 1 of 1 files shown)


sidecar/nginx/lua/amalgam8.lua
```diff
@@ -691,6 +686,18 @@ function Amalgam8:apply_rules()
    local selected_backend = nil
    local cookie_version = ngx.var.cookie_version --check for version cookie
 
+   -- if we have routes/actions for the destination but no instances, then return HTTP 503
+   -- else return HTTP 404, as we don't know if this service actually exists or not.
+
+   if not instances or #instances == 0 then
+      if routes or actions then
+         ngx.status = ngx.HTTP_SERVICE_UNAVAILABLE
```
GregHanson, line 694: I follow that you want to distinguish between no instances found vs no service found - but do we really want to through a 503 error code?  Aren't 5xx codes typically server faults?  If the user hasn't created any instances for that service - is it the proxy's fault?

rshriram: 500 is an internal server error. 503 is typically used, afaik, to indicate that the service is unable to serve any requests, because its overloaded/unavailable/disabled/etc., but it also indicates to the caller that the service actually exists.

https://www.microsoft.com/technet/prodtechnol/windowsserver2003/library/iis/55f71614-ef1b-4015-b9c8-a42c1e700c25.mspx?mfr=true

However, 404 indicates that the proxy has no idea about the service being requested.

The reason I converted from 404 to 503, is because people distinguish between a HTTP resource that exists but is unreachable vs a resource that does not exist. If it does not exist (404), they try to create a resource (e.g., create a user, if a 404 is returned by login service) vs return a "try again later" if 503 is returned by login service.

Now that we have health checks, if all instances of the service go down due to say network issues, returning a 404 causes confusion, as the service exists, but we can't reach it. A 503 in this scenario is much more representative.

GregHanson (approved):
LGTM

PR #344, opened 2016-10-14
build sidecar version of examples without alpine

rshriram:
all python images are based on python-2, which is not alpine based. Removing v1-alpine from example build scripts.


GregHanson (approved):
LGTM

PR #347, opened 2016-10-18
deregisters with registry when app terminates

GregHanson:
also removed filebeat


rshriram (changes requested):
Lets not remove the filebeat code until we fix #52 . Otherwise, all examples will be broken (Gremlin part atleast).

rshriram (approved):

rshriram:
fixes #50

PR #349, opened 2016-10-19
configured supervisor to manage multiple processes

GregHanson:
no longer starts/stops just app and filebeat.  The new app supervisor can be configured to start/stop any number of processes


sidecar/supervisor/appsupervisor.go
```diff
@@ -28,33 +28,46 @@ import (
...
 type AppSupervisor struct {
 	agent *register.RegistrationAgent
+	cmds  []*exec.Cmd
+}
+
+// AddProcess TODO
+func (a *AppSupervisor) AddProcess(cmdArgs, env []string) {
+	cmd := exec.Command(cmdArgs[0], cmdArgs[1:]...)
+	if a.cmds == nil {
+		a.cmds = make([]*exec.Cmd, 0)
```
ijsnellf: I think the suggested style is: `a.cmds = []*exec.Cmd{}`

sidecar/sidecar.go
```diff
@@ -110,8 +110,7 @@ func Run(conf config.Config) error {
 			return err
 		}
 
-		// TODO: Log failure?
-		go appSupervisor.DoLogManagement("/tmp/filebeat.yml")
+		appSupervisor.AddProcess([]string{"filebeat", "-c", "/tmp/filebeat.yml"}, []string{"GODEBUG=netdns=go"})
```
rshriram: Can we get rid of this completely?
Basically a YAML file format like this

``` yaml
supervise:
 app: "cmd string with args"
 helpers:
   - helper cmd 1..
   - helper cmd 2..
```

We kill the container only if the app dies. if the helpers die, we don't do anything (for now). In future, we may want to restart it automatically.

sidecar/supervisor/appsupervisor.go
```diff
@@ -64,11 +77,7 @@ func (a *AppSupervisor) DoAppSupervision(cmdArgs []string, agent *register.Regis
 		select {
 		case sig := <-sigChan:
 			log.Infof("Intercepted signal '%s'", sig)
-			if appProcess != nil {
-				appProcess.Signal(sig)
-			} else {
-				a.Shutdown(0)
-			}
+			a.Shutdown(0)
```
rshriram, line 110: Need to check if its the app process or a helper process. Only the app process dies, then we kill all other processes.

The current implementation will kill the container even if a helper process like filebeat dies.

rshriram (changes requested):
one bug fix and a YAML mod needed to make this more generic. Along with this mod, we should migrate our examples to use this model.
And test cases :)

sidecar/supervisor/appsupervisor.go
```diff
@@ -94,8 +103,12 @@ func (a *AppSupervisor) DoAppSupervision(cmdArgs []string, agent *register.Regis
 func (a *AppSupervisor) Shutdown(exitCode int) {
 	// TODO: Gracefully shutdown Ngnix, and filebeat
 	//ugly temporary hack to kill all processes in container
-	exec.Command("pkill", "-9", "-f", "nginx")
-	exec.Command("pkill", "-9", "-f", "filebeat")
+	//exec.Command("pkill", "-9", "-f", "nginx")
+	//exec.Command("pkill", "-9", "-f", "filebeat")
+
+	for _, cmd := range a.cmds {
+		cmd.Process.Kill()
```
zcahana: In case of SIGTERM/SIGINT, we should relay the signal to child processes, and allow them gracefully shutdown.
This can take some time to complete, so should better be done with a separate goroutine per process (can use a `sync.WaitGroup` to wait for all to finish/timeout).

sidecar/supervisor/appsupervisor.go
```diff
@@ -94,8 +103,12 @@ func (a *AppSupervisor) DoAppSupervision(cmdArgs []string, agent *register.Regis
...
+	}
 
 	if a.agent != nil {
 		a.agent.Stop()
```
zcahana, line 147: There seem to be a bug here (from #347). After the agent has been stopped, it might be immediately restarted by the healthchecker. An optional solution is to not pass in `RegistrationAgent`, but rather pass in some `Startable` interface, which is implemented by both the `RegistrationAgent` and healthchecker.

zcahana, line 147: There's also several items missing from this change-set:
1. Need to properly manage Nginx process(es).
2. Need to provide configuration option for multiple processes (via YAML file).
3. Given that the sidecar is expected to run as [PID 1](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/) in the container, we need to properly reap adopted child processes.

zcahana (changes requested):

sidecar/sidecar.go
GregHanson: @rshriram are we getting rid of the command arg log flag completely then?  How would someone supervise an app and log?  Two `apps` provided in the yaml file, one for logging and one for the app registered by service discovery?  Because if the logger is always a `helper` command, how would you log in this case?

rshriram: Yes, we would get rid of the -log and -logstash_host flags completely. From our perspective, wouldn't care what the helper process is. The only semantics we offer is that termination of a helper process does not result in termination of the sidecar itself. But termination of the primary process will result in termination of the sidecar.

``` yaml
supervise:
  app: python /opt/microservices/productpage.py 9080 http://localhost:6379
  helpers:
     - filebeat -c /etc/filebeat.yaml
     - statsd -x -y -z -a -b
     - logrotate -a -b -c
```

sidecar/supervisor/appsupervisor.go
ijsnellf: NGINX was not being managed prior to this PR. I agree we should manage it, but it might make sense to do that as part of its own PR?

zcahana: Nginx management can go in another PR.

sidecar/sidecar.go
```diff
@@ -111,7 +112,8 @@ func Run(conf config.Config) error {
 			return err
 		}
 
-		appSupervisor.AddProcess([]string{"filebeat", "-c", "/tmp/filebeat.yml"}, []string{"GODEBUG=netdns=go"})
+		// TODO: Log failure?
+		go appSupervisor.DoLogManagement("/tmp/filebeat.yml")
```
rshriram: We don't need this anymore.

sidecar/supervisor/appsupervisor.go
```diff
@@ -21,52 +21,73 @@ import (
...
-		log.Infof("Launching app '%s' with args '%s'", cmd.Args[0], strings.Join(cmd.Args[1:], " "))
+	appChan := make(chan error, 1)
+	if a.App != nil {
+		log.Infof("Launching app '%s' with args '%s'", a.App.Args[0], strings.Join(a.App.Args[1:], " "))
+		err := a.App.Start()
+		if err != nil {
+			appChan <- err
+		} else {
+			go func() {
+				appChan <- a.App.Wait()
```
ijsnellf: Do we have to do a `.Wait()` if `.Start()` fails? I thought `.Start()` failing implied that the process was not started.

sidecar/supervisor/appsupervisor.go
```diff
@@ -21,52 +21,73 @@ import (
...
+	"github.com/amalgam8/amalgam8/sidecar/config"
 	"github.com/amalgam8/amalgam8/sidecar/register"
 )
 
 // AppSupervisor TODO
 type AppSupervisor struct {
-	agent *register.RegistrationAgent
-	cmds  []*exec.Cmd
+	agent   *register.RegistrationAgent
+	App     *exec.Cmd
```
ijsnellf: Is there a reason these are public?

sidecar/supervisor/appsupervisor.go
```diff
@@ -21,52 +21,73 @@ import (
...
+		if cmd.Env != nil {
+			cmdEnv = append(cmdEnv, cmd.Env...)
+		}
+		osCmd.Env = cmdEnv
 
-	cmdEnv := os.Environ()
-	if env != nil {
-		cmdEnv = append(cmdEnv, env...)
+		if cmd.Primary {
+			a.App = osCmd
```
ijsnellf: We only support one primary app, but you can specify multiple apps are primary? Only the last command specified as primary will run as primary, which seems misleading.

sidecar/supervisor/appsupervisor.go
```diff
@@ -21,52 +21,73 @@ import (
...
 	log "github.com/Sirupsen/logrus"
+	"github.com/amalgam8/amalgam8/sidecar/config"
 	"github.com/amalgam8/amalgam8/sidecar/register"
 )
 
 // AppSupervisor TODO
 type AppSupervisor struct {
-	agent *register.RegistrationAgent
-	cmds  []*exec.Cmd
+	agent   *register.RegistrationAgent
```
ijsnellf: Maybe not in this PR --but, we should switch to a `Start()`, `Stop()` interface that can be fulfilled by `*register.RegistrationAgent` and `*register.HealthChecker` so that we can ensure that we actually deregister and don't re-register.

sidecar/sidecar.go
ijsnellf: We can't remove it without making a bunch of other modifications so that we don't break the examples. I feel like it can be a separate PR.

sidecar/config/config.go
```diff
@@ -28,11 +28,17 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+const (
+	RestartOnFailure   = "restart"
+	KillOnFailure      = "kill"
+	DoNothingOnFailrue = "nothing"
```
ijsnellf: Typo.

sidecar/config/config.go
```diff
@@ -28,11 +28,17 @@ import (
...
+)
+
 // Command TODO
```
ijsnellf: Add actual comments instead of `TODO`?

ijsnellf (commented):
Minor comments.

sidecar/config/config.go
```diff
@@ -204,7 +213,11 @@ func (c *Config) loadFromContext(context *cli.Context) error {
 	}
 
 	if context.Args().Present() {
-		c.App = context.Args()
+		cmd := Command{
+			Cmd:     context.Args(),
+			Primary: true,
+		}
+		c.Commands = []Command{cmd}
```
zcahana: `c.Commands = []Command{cmd}`
Maybe append instead of overwrite? This is also the approach we've been taking with the healthcheck array.

sidecar/supervisor/appsupervisor.go
```diff
@@ -21,52 +21,73 @@ import (
...
+		Helpers: []*exec.Cmd{},
 	}
 
-	cmd.Stdin = os.Stdin
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	for _, cmd := range conf.Commands {
+		osCmd := exec.Command(cmd.Cmd[0], cmd.Cmd[1:]...)
+
+		osCmd.Stdin = os.Stdin
```
zcahana: Mapping stdin to multiple processes doesn't make much sense. If the need arises, we can add a configuration to map stdin to an individual process.

sidecar/supervisor/appsupervisor.go
```diff
@@ -21,52 +21,73 @@ import (
...
+		osCmd.Stdout = os.Stdout
+		osCmd.Stderr = os.Stderr
+
+		cmdEnv := os.Environ()
+		if cmd.Env != nil {
+			cmdEnv = append(cmdEnv, cmd.Env...)
```
zcahana: Can replace lines 52-56 with a one-liner:

``` go
osCmd.Env = append(os.Environ(), cmd.Env...)
```

Actually, if `cmd.Env` is empty/nil, we can avoid setting `osCmd.Env` alltogether (not a significant optimization though):

``` go
if len(cmd.Env) > 0 {
    osCmd.Env = append(os.Environ(), cmd.Env...)
}
```

sidecar/supervisor/appsupervisor.go
zcahana: The golang docs aren't very clear on that, but it seems unharmful to `Wait()` in any case.

sidecar/supervisor/appsupervisor.go
```diff
@@ -21,52 +21,73 @@ import (
...
+		log.Infof("Launching app '%s' with args '%s'", cmd.Args[0], strings.Join(cmd.Args[1:], " "))
 		go func(cmd *exec.Cmd) {
-			err := cmd.Start()
-			if err != nil {
-				appChan <- err
-			} else {
-				appChan <- cmd.Wait()
-			}
+			err := cmd.Run()
+			log.WithError(err).Warn("Failed to launch helper command '%s'", cmd.Args[0])
```
zcahana: `cmd.Run()` will return nil in case of normal termination (code 0).

sidecar/supervisor/appsupervisor.go
```diff
@@ -77,6 +98,10 @@ func (a *AppSupervisor) DoAppSupervision(agent *register.RegistrationAgent) {
 		select {
 		case sig := <-sigChan:
 			log.Infof("Intercepted signal '%s'", sig)
+
+			// forwarding signal to application parent process
+			exit(append(a.Helpers, a.App), sig)
```
zcahana: `append(a.Helpers, a.App)` is not a very good idea, because it will most likely modify `a.Helpers` instead of reallocating (assuming its capacity > len [most likely]).

Also, the use of `exit(cmds, sig)` here is a bit fragile, as it might be called multiple times, possibly concurrently (e.g.., two consecutive signals or a signal followed by appChan error). Currently, looks like the only consequence is multiple calls to `Wait()` for the same process... do you see any other problem with it?

Also, the naming of `exit()` alongside `Shutdown()` is somewhat confusing - maybe rename `exit()` to `terminateSubprocesses()` or the like?

sidecar/supervisor/appsupervisor.go
```diff
@@ -93,29 +118,42 @@ func (a *AppSupervisor) DoAppSupervision(agent *register.RegistrationAgent) {
 					log.Errorf("App failed to start: %v", err)
 				}
 			}
-
+			exit(a.Helpers, syscall.SIGKILL)
```
zcahana: It is better to try and gracefully terminate (SIGTERM) the helpers first before killing them.

sidecar/supervisor/appsupervisor.go
```diff
@@ -118,8 +129,23 @@ func (a *AppSupervisor) DoAppSupervision(agent *register.RegistrationAgent) {
...
-			a.Shutdown(exitCode)
+
+			switch err.Proc.Action {
+			case config.DoNothingOnFailrue:
+				//Ignore this dead process
+				log.WithError(err.Err).Warn("App '%s' with args '%s' exited with error.  Ignoring", err.Proc.Cmd.Args[0], strings.Join(err.Proc.Cmd.Args[1:], " "))
+				err.Proc.Cmd.Wait()
+
+			case config.RestartOnFailure:
+				//TODO add restart functionality
```
zcahana: If restart is not implemented, I'd suggest to remove it from the configuration, and open another issue to implement it.

sidecar/config/config.go
```diff
@@ -28,11 +28,17 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+const (
```
zcahana, line 33: How about: `on_exit` instead of `on_death`, and `restart`, `ignore`, and `terminate` as values?

sidecar/sidecar.go
```diff
@@ -93,29 +93,6 @@ func Run(conf config.Config) error {
 
 	appSupervisor := supervisor.NewAppSupervisor(&conf)
 
-	if conf.Log {
-		//Replace the LOGSTASH_REPLACEME string in filebeat.yml with
-		//the value provided by the user
-		//TODO: Make this configurable
-		filebeatConf := "/etc/filebeat/filebeat.yml"
```
zcahana, line 100: Also can delete the `/docker/filebeat.yml` file from our repo, and remove its references from the Dockerfiles.

sidecar/supervisor/appsupervisor.go
ijsnellf: I think `append` does the nil check internally.

zcahana: It does. The idea was that if `cmd.Env` is nil/empty, we leave `osCmd.Env` nil as well, and save some strings processing, etc when starting the process.

sidecar/sidecar.go
GregHanson: this will have to be done in a separate PR since gremlin still uses this in the examples/ testing

sidecar/supervisor/appsupervisor.go
ijsnellf: In the code `exit(cmds, sig)`  is followed by `Shutdown(exitCode)`, so `exit` won't ever be called more than once, since `Shutdown` blocks until the sidecar process terminates.

sidecar/config/config_test.go
```diff
@@ -320,8 +301,6 @@ log_level: debug
 			Expect(c.HealthChecks[1].Timeout).To(Equal(time.Duration(3) * time.Second))
 			Expect(c.HealthChecks[1].Method).To(Equal("POST"))
 			Expect(c.HealthChecks[1].Code).To(Equal(201))
-			Expect(c.Log).To(Equal(true))
-			Expect(c.LogstashServer).To(Equal("logstash:8092"))
 			Expect(c.LogLevel).To(Equal("debug"))
```
rshriram, line 306: I am hoping you will add test cases, after this series?

sidecar/sidecar.go
rshriram: So this is gone right? Because the diff shows that its being removed. Its needed only in examples, not testing. And if the examples work correctly, it will validate that this subsystem works properly. 

The changes to be made to examples will be in the form of adding a sidecar.yml config file, where we add a shell script that would generate a filebeat.yaml config and launch filebeat.

sidecar/sidecar.go
```diff
@@ -172,11 +150,7 @@ func Run(conf config.Config) error {
 		}
 	}
 
-	if conf.Supervise {
-		appSupervisor.DoAppSupervision(conf.App, registrationAgent)
-	} else {
-		select {}
-	}
+	appSupervisor.DoAppSupervision(registrationAgent)
```
rshriram, line 151: Why did you remove this? If there was no supervision, then you wouldn't need one right?
Specifically, how about this:

```
if conf.Supervise {
        appSupervisor := supervisor.NewAppSupervisor(&conf)
    appSupervisor.DoAppSupervision(conf.App, registrationAgent)
} else {
    select {}
}
```

sidecar/supervisor/appsupervisor.go
```diff
@@ -21,39 +21,79 @@ import (
...
-	agent *register.RegistrationAgent
+	agent     *register.RegistrationAgent
+	processes []*process
 }
 
-// DoAppSupervision TODO
-func (a *AppSupervisor) DoAppSupervision(cmdArgs []string, agent *register.RegistrationAgent) {
+type process struct {
+	Cmd    *exec.Cmd
+	Action string
```
rshriram, line 39: rename to OnExit ?

sidecar/supervisor/appsupervisor.go
```diff
@@ -85,38 +125,50 @@ func (a *AppSupervisor) DoAppSupervision(cmdArgs []string, agent *register.Regis
...
-	logcmd.Stdin = os.Stdin
-	logcmd.Stdout = os.Stdout
-	err := logcmd.Run()
-	if err != nil {
-		log.WithError(err).Warn("Filebeat process exited with error")
+func terminateSubprocesses(procs []*process, sig os.Signal) {
+
+	var wg sync.WaitGroup
+
+	wg.Add(len(procs))
```
rshriram: Check for len 0 ?

rshriram (changes requested):

sidecar/supervisor/appsupervisor.go
GregHanson: not needed, this works just fine:

```
    var wg sync.WaitGroup

    procs := []int{}

    wg.Add(len(procs))
    for _, i := range procs {
        defer wg.Done()
        fmt.Println("Infinite loop ", i)
        time.Sleep(time.Second*5)
    }

    wg.Wait()
    fmt.Println("DONE!")
```

sidecar/supervisor/appsupervisor.go
zcahana: You're right, my mistake :smile:

sidecar/sidecar.go
zcahana: The reason is that `DoAppSupervision()` does more than just supervising child processes: it also intercepts signals (e.g. SIGTERM due to `docker stop`), and stops the registration agent (so that it can deregister).

zcahana (approved):

rshriram (approved):

Issue #314, opened 2016-10-02
Implement additional healthcheck types


zcahana:
Currently we have support for HTTP-based health checks only. It can be useful to support several other types:
- TCP-based healthcheck - attempt to establish a TCP connection to a given `host:port`.
- Command-based healthcheck - invoke a given command (optionally with arguments) and examine exit status.

Example configuration:

``` yaml
healthchecks:
  - type: tcp
    value: localhost:5053
    interval: 30s
    timeout: 1s

  - type: command
    value: /opt/myapp/health.sh
    interval: 60s
    timeout: 3s

  - type: command
    value: /bin/grep
    args: [ "-i", "ERROR", "/opt/myapp/out.log"]
    interval: 60s
    timeout: 3s
```

Note: it may be beneficial to refactor the healthcheck mechanism so that `HealthCheck` interface is responsible for a single execution of the healthcheck only (i.e., `Execute()` instead of `Start()` and `Stop()`).  This way we don't have to re-implement the scheduling logic in each of the healthcheck types. This also make it easier to test each of the healthcheck implementations.

PR #350, opened 2016-10-20
TCP and Command health checks

ijsnellf:
This PR refactors some of the health check logic and adds TCP and Command health checks as discussed in https://github.com/amalgam8/amalgam8/issues/314.

The health check implementation has been refactored by separating the periodic execution of health checks from the execution of a single health check. Now the healthcheck.Check interface provides a single `Execute` method

``` go
// Check is an interface for performing a health check.
type Check interface {
    // Execute the health check and return an error on failure.
    Execute() error
}
```

and the periodic execution of a health check is handled by `Agent`, which takes in a `Check` and an interval and controls the periodic execution of the health check via `Start` and `Stop` methods.


ijsnellf:
This PR is a first draft. Additional automated tests should be added, but I wanted to get feedback on overall my approach before adding all the tests.

sidecar/config/config.go
```diff
@@ -215,6 +217,8 @@ func (c *Config) loadFromContext(context *cli.Context) error {
 				hcType = HTTPHealthCheck
 			case "https":
 				hcType = HTTPSHealthCheck
+			case "tcp":
+				hcType = TCPHealthCheck
```
zcahana, line 221: Missing a `case "cmd":` here.

sidecar/register/healthcheck/http.go
```diff
@@ -0,0 +1,132 @@
...
+type HTTP struct {
+	client *http.Client
+
+	url    string
+	method string
+	code   int
+}
+
+// NewHTTPAgent creates a new HTTP health check.
+func NewHTTPAgent(conf config.HealthCheck) (*Agent, error) {
```
zcahana: The `Agent` is check-agonstic, so there's no need for `NewFooAgent` for each check type.
We can have a `NewHTTPCheck(conf)` constructor, and let the `BuildAgent()` function wrap it with a new `Agent`.

sidecar/register/healthchecker.go
```diff
@@ -71,34 +72,35 @@ func (checker *HealthChecker) Stop() {
...
 
-	statusChan := make(chan HealthStatus, len(checker.checks))
+	statusChan := make(chan healthcheck.Status, len(checker.checks))
 	for _, healthCheck := range checker.checks {
 		healthCheck.Start(statusChan)
 	}
 
 	// Set of health checks that have most recently reported unhealthy statuses.
-	unhealthyStatuses := make(map[HealthCheck]interface{})
+	unhealthyStatuses := make(map[healthcheck.Check]interface{})
```
zcahana: Having `healthcheck.Check` as the map key is problematic, because map keys must be [comparable](https://golang.org/ref/spec#Comparison_operators). Particularly, `healthcheck.Command` is not comparable because it has a slice field (`Args []string`) which isn't comparable, so this would have panicked at runtime.

We can make sure that each healthcheck is comparable, but that imposes some unjustified restrictions on the healthchecks.

Other solution would be to dismiss the `unhealthyStatuses` set in favor of a simple counter of how many healthchecks are currently healthy. This requires each healthcheck to have a channel of its own. Something along this line:

``` go
healthyChan := make(chan bool, len(checker.checks))

for _, healthCheck := range checker.checks {
    statusChan := make(chan healthcheck.Status, 1)
    healthCheck.Start(statusChan)

    go func() {
        wasHealthy := false
        for {
            select {
            case status <- statusChan:
                healthy := status.Error == nil
                if healthy != wasHealthy {
                    healthyChan <- healthy
                }
                wasHealthy = healthy
            }
        }
    }()
}

numHealthy := 0
for {
    select {
    case healthy <- healthyChan:
        if healthy {
            numHealthy++
            if numHealthy == len(checker.checks) {
                checker.registration.Start()
            }
        } else {
            numHealthy--
            if numHealthy == len(checker.checks) - 1 {
                checker.registration.Stop()
            }
        }
    }
}
```

zcahana (changes requested):
Overall this looks good, but please see comment on collecting/coalescing statuses from different healthchecks.

sidecar/register/healthchecker.go
ijsnellf: According the Golang docs:

> Interface values are comparable. Two interface values are equal if they have identical dynamic types and equal dynamic values or if both have value nil.

The dynamic value is a pointer (in this case, `*healthcheck.HTTP` or similar). 

For Golang maps:

> The comparison operators == and != must be fully defined for operands of the key type [...]

Which pointers fulfill:

> Pointer values are comparable. Two pointer values are equal if they point to the same variable or if both have value nil. Pointers to distinct zero-size variables may or may not be equal.

Since pointers are being compared and not the values, there should not be any issues with this implementation.

edit:
You are correct if we ever wanted to create an implementation of `healthcheck.Check` that uses non-pointer receivers.

``` golang
type Example struct {
   arr []int
}

func (e Example) Execute() error {
   // Do stuff
}
```

This would cause panics as you described. We don't do that anywhere currently, but I can agree that is a liability.

sidecar/register/healthcheck/http.go
ijsnellf: Makes sense to me. I was on the fence between doing it that way or the way I implemented it, but you make an excellent point.

sidecar/register/healthchecker.go
ijsnellf: To avoid the liability, I switched to an implementation of the health checker similar to your suggestion.

zcahana: You're right, I was missing the point that these are the pointers that need to be comparable.

sidecar/config/config.go
```diff
@@ -219,14 +219,26 @@ func (c *Config) loadFromContext(context *cli.Context) error {
 				hcType = HTTPSHealthCheck
 			case "tcp":
 				hcType = TCPHealthCheck
+			case "cmd":
```
zcahana: I'm wondering if the standardized [file URI scheme](https://en.wikipedia.org/wiki/File_URI_scheme) (e.g., `file:///usr/bin/health.sh`) could be a better fit here.

zcahana (approved):
Only comment is to consider `file:///` instead of `cmd:///` (yes, it contradicts my prev comment), but I leave the decision to you. Otherwise this LGTM.

GregHanson:
resolves issue #314

PR #355, opened 2016-10-25
Updated dependencies

ijsnellf:
Removed the Sarama library that is no longer used and locked down the version of Redigo.


GregHanson (approved):
LGTM

Issue #339, opened 2016-10-13
Need more informative messages for user


dougbeauchene:
I created a very simple container and integrated the sidecar with it.

When I start the container I see these messages logged in the container logs:

ctime="2016-10-13T16:24:39Z" level=info msg="No hostname is configured. Using local IP instead..." 
Ltime="2016-10-13T16:24:39Z" level=info msg="Obtained local IP 172.32.0.74" 
8time="2016-10-13T16:24:39Z" level=info msg=Registering 
=time="2016-10-13T16:24:39Z" level=info msg="Starting NGINX" 

I am guessing that the "Registering" and "Starting NGINX" messages are coming from the sidecar logic.

I would suggest updating these messages, and other messages the sidecar is generating to be more informative for the user.  I think it will give the user a better idea of what is happening, progress of various parts of their microservice solution, etc.

For example...
"Registering with Amalgam8 registry"
"Starting NGINX in the Amalgam8 sidecar"

Yes it's more wordy, but it does help the user easily and quickly assess success/failure.  

I'd also suggest adding some of this info to the documentation so the user can do some verification as they stand up their microservice solution using Amalgam8.

ijsnellf:
We have to strike a balance between overly verbose logging, which can make things impossible to read with too much redundant information, and excessively sparse logging which doesn't provide proper context.

In general, I think references to Amalgam8 should be omitted inside the Amalgam8 logs. Unless otherwise stated, it should be implicit that we are communicating with the Amalgam8 registry, that Amalgam8 health checking agent started, etc.

However, I agree we could provide better context. My suggestion for this is to adopt a similar logging scheme to registry. Registry provides a field in the logs that describes the module the log entry was generated in. That should provide proper context without adding unnecessary information into the log messages themselves. This suggestion is tracked in #297.

ijsnellf:
Since I think this will be handled by #297, I am closing this issue.

PR #359, opened 2016-10-27
Fix registration agent to synchronously deregister when stopped

zcahana:
This PR fixes a bug in the sidecar shutdown logic where the service deregistration attempt is performed asynchronously with a background goroutine, while the main goroutine already exits. This causes the deregistration goroutine to exit as well, most likely before it managed to actually deregister.

Additionally, this PR adds unit tests for the basic regisration agent functions, as well as improved logging (which partially resolves issue #339).

### DIFF
--- a/sidecar/register/registration.go
+++ b/sidecar/register/registration.go
@@ -57,6 +57,9 @@ func NewRegistrationAgent(config RegistrationConfig) (*RegistrationAgent, error)
 // Start maintaining registration with registry.
 // Non-blocking.
 func (agent *RegistrationAgent) Start() {
+	logrus.WithField("service_name", agent.config.ServiceInstance.ServiceName).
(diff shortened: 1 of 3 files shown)


ijsnellf:
Looks good.

An additional item is to change the `*RegistrationAgent` reference in `appsupervisor.go` to a simple `Start()`/`Stop()` interface and then pass in `HealthChecker` instead, so that there is no chance of `HealthChecker` trying to re-register before the process is shut down.

See this line:
https://github.com/amalgam8/amalgam8/blob/cd3d671821684aa2db482988458c8b3964aecc73/sidecar/supervisor/appsupervisor.go#L30

I think that can be handled as part of https://github.com/amalgam8/amalgam8/pull/349.

ijsnellf (approved):
LGTM.

PR #362, opened 2016-10-27
Logging of source IP address in Sidecar routing requests

CarltonSemple:
I added the source IP address for each request to the logs. This updates repeatedly so that if the sidecar's IP address changes, it should show the current IP address.  
The information will be make the logs more useful since you will now be able to see the information for each connection from Source A to Destination B along with a firm way to identify which container is generating Source A

image


sidecar/nginx/lua/amalgam8.lua
```diff
@@ -476,6 +478,7 @@ function Amalgam8:new()
 
    self.myname = myname
    self.mytags = mytags
+   self.myip = "0.0.0.0"
```
rshriram: Why can't you fetch the IP here?
The IP address of the container will not change once launched. The init function is executed when nginx is (re)started in the container.

docker/Dockerfile.sidecar.ubuntu
```diff
@@ -32,6 +32,8 @@ RUN set -ex \
     \
     && rm -rf /tmp/filebeat*
 
+RUN apt-get update; apt-get upgrade -y
+RUN apt-get install -y net-tools
```
rshriram: You don't need to install anything. IP address can be obtained in a simpler way, as shown below.

sidecar/nginx/lua/amalgam8.lua
```diff
@@ -498,6 +501,8 @@ function Amalgam8:update_state(input)
    local a8_actions = {}
    local err
 
+   self.myip = self:os_cmd("ifconfig eth0 | grep 'inet addr:' | cut -d: -f2 | awk '{ print $1}'", false)
```
rshriram: Move to init block? Doing this in update_state simply adds overhead.

Secondly, to get the local IP

``` bash
hostname -i
```

sidecar/nginx/lua/amalgam8.lua
```diff
@@ -852,5 +857,24 @@ function Amalgam8:get_mytags()
    return self.mytags
 end
 
+function Amalgam8:get_myip()
+   return self.myip
+end
+
+function Amalgam8:get_env(name)
```
rshriram: This function seems unnecessary. Not used anywhere.

sidecar/nginx/lua/amalgam8.lua
```diff
@@ -852,5 +857,24 @@ function Amalgam8:get_mytags()
...
+   return os.getenv(name)
+end
+
+function Amalgam8:os_cmd(cmd, raw)
```
rshriram: Given that you are using this only to parse the IP address, you might as well rename this function to something like ip_cmd() and you just have to return the result of the "hostname -i" without the new lines.

rshriram (changes requested):

sidecar/nginx/lua/amalgam8.lua
CarltonSemple: i had moved it to this position for the possibility that the IP address might change. i can move it back though

CarltonSemple: i tried `hostname -i` on my VM. it gave me two responses, but in one of the containers it gave me one IP

docker/Dockerfile.sidecar.ubuntu
CarltonSemple: alrighty, will do

sidecar/nginx/lua/amalgam8.lua
CarltonSemple: oh, i see.

rshriram (approved):

CarltonSemple:
i checked `hostname -i` with a container i have on bluemix, and it was returning two of the same IP address, so that's why I added a filter to the command

Issue #361, opened 2016-10-27
Revision number not updating with in-memory backend during rule PUT


ijsnellf:
The revision number is not updated on a `PUT` to `/v1/rules`.

``` bash
dev@dev:~/Desktop/Workspace/go/src/github.com/amalgam8/amalgam8/examples$ curl -i  http://dev-jem-a8-controller.mybluemix.net/v1/rulesHTTP/1.1 200 OK
X-Backside-Transport: OK OK
Connection: Keep-Alive
Transfer-Encoding: chunked
A8-Request-Id: 34bc1bfb-6a80-49fa-83b1-e5d837bf5951
Content-Type: application/json; charset=utf-8
Date: Wed, 26 Oct 2016 22:09:41 GMT
X-Global-Transaction-ID: 1912894413

{"rules":[{"id":"a11437a5-1613-43e4-bb0b-e5e655e528fb","priority":1,"destination":"productpage","route":{"backends":[{"tags":["v1"]}]}},{"id":"0bd7f642-3fba-4fe0-b9b3-f24c7aa7c5ae","priority":1,"destination":"ratings","route":{"backends":[{"tags":["v1"]}]}},{"id":"b8d3f12f-08d1-451b-a3fd-38a86bb0952c","priority":1,"destination":"details","route":{"backends":[{"tags":["v1"]}]}},{"id":"4247b6cf-a3ed-4a30-aaaa-a1f198e741b1","priority":1,"destination":"reviews","route":{"backends":[{"weight":0.5,"tags":["v2"]},{"tags":["v3"]}]}}],"revision":19}dev@dev:~/Desktop/Workspace/go/src/github.com/amalgam8/amalgam8/examples$ 
dev@dev:~/Desktop/Workspace/go/src/github.com/amalgam8/amalgam8/examples$ 
dev@dev:~/Desktop/Workspace/go/src/github.com/amalgam8/amalgam8/examples$ curl -i -X PUT http://dev-jem-a8-controller.mybluemix.net/v1/rules  -H 'Content-type: application/json' -H 'Accept: application/json' --data '{"rules": [{"priority": 1, "route": {"backends": [{"weight": 0.6, "tags": ["v2"]}, {"tags": ["v3"]}]}, "destination": "reviews", "id": "4247b6cf-a3ed-4a30-aaaa-a1f198e741b1"}]}'
HTTP/1.1 200 OK
X-Backside-Transport: OK OK
Connection: Keep-Alive
[...]

Issue #354, opened 2016-10-25
More unit tests for controller and sidecar code


ijsnellf:
Controller and sidecar have very low UT coverage. An effort needs to be made to improve the level of coverage.

PR #363, opened 2016-10-27
Fix rule update bug

ijsnellf:
Fixed bug where revision was not incrementing on rule updates and added some UTs to check for similar issues.


ijsnellf:
Resolves https://github.com/amalgam8/amalgam8/issues/361.

rshriram (approved):

ijsnellf:
Contributes to #354.

PR #365, opened 2016-10-28
Added service TTL config option

### LINKED ISSUE #360: Add support for configurable service TTL
This will allow to tune the sidecar for better balance between expiration time and heartbeats load:
- Lower TTL: Faster expiration (and failover), more heartbeats load
- Higher TTL: Slower expiration, less heartbeat load

Additionally, we can have the heartbeats frequency be configurable as well, to further balance between heartbeats load and false expiration due to missed heartbeats:
- Lower frequency: less heartbeats load, more false expirations
- Higher frequency: more heartbeats load, less false expirations

...or we can just default the frequency to be TTL/2 or TTL/3 (as is now).

ijsnellf:
Resolves #360.


ijsnellf:
I have TTL default to 60s right now.

sidecar/config/config.go
```diff
@@ -38,9 +38,10 @@ type Service struct {
 
 // Endpoint configuration
 type Endpoint struct {
-	Host string `yaml:"host"`
-	Port int    `yaml:"port"`
-	Type string `yaml:"type"`
+	Host string        `yaml:"host"`
+	Port int           `yaml:"port"`
+	Type string        `yaml:"type"`
+	TTL  time.Duration `yaml:"ttl"`
```
zcahana, line 59: I'm not sure endpoint is the right place for the TTL. In the Registry's API, the TTL and endpoint are both attributes of the registered service instance.

Perhaps we can have the TTL (and later on, hearbeat frequency, maybe) under the `register` attribute:

``` yaml
register:
    ttl: 60s
    heartbeat: 20s
```

And ideally, still support the boolean `register` flag that uses default values for ttl/heartbeat:

``` yaml
register: true
```

sidecar/config/config.go
```diff
@@ -287,6 +289,7 @@ func (c *Config) Validate() error {
 			IsNotEmpty("Service Name", c.Service.Name),
 			IsInRange("Service Endpoint Port", c.Endpoint.Port, 1, 65535),
 			IsInSet("Service Endpoint Type", c.Endpoint.Type, []string{"http", "https", "tcp", "udp", "user"}),
+			IsInRangeDuration("Service Endpoint TTL", c.Endpoint.TTL, 30*time.Second, 120*time.Second),
```
zcahana: I don't see a reason to restrict the TTL that much. How about [10s, 10m]?

zcahana (changes requested):

sidecar/config/config.go
ijsnellf: Has registry been modified to support TTLs less than 30s?

zcahana: The registry supports configuration of [min,max] TTL, which defaults to [10s, 10m].

sidecar/config/config.go
rshriram: @ijsnellf is this supported yet?

ijsnellf: No.

rshriram:
should we can this for the moment?

PR #366, opened 2016-10-30
Fix the superviser to stop the healthchecker upon termination

zcahana:
Fixes the problem highlighted [here](https://github.com/amalgam8/amalgam8/pull/349#discussion_r84232079) and [here](https://github.com/amalgam8/amalgam8/pull/359#issuecomment-256735648).


GregHanson (approved):
LGTM

PR #368, opened 2016-10-31
Fixed bug resulting in YAML errors not being returned

ijsnellf:
This bug resulted in YAML config parsing failing silently.


zcahana (approved):
Good catch. LGTM.

PR #370, opened 2016-11-01
add retry and timeout support in nginx

rshriram:
(re) add support for retries and timeouts in the balancer_by_lua logic. This was removed in 0.4.0 due to the inability to resolve DNS names in balancer_by_lua context. The problem still remains, however the retry and timeout logic is essential for automatically failing over to other instances in the LB pool when the current instance has died (but whose TTL has not expired in the registry). refer to #50 and #359.


sidecar/nginx/lua/amalgam8.lua
```diff
@@ -787,11 +787,14 @@ function Amalgam8:apply_rules()
...
-   ngx.var.a8_upstream_tags = upstream.tags
+   ngx.ctx.a8_upstreams = selected_instances
+   ngx.var.a8_upstream_tags = ""
+   if selected_backend then
+      ngx.ctx.a8_timeout = selected_backend.timeout
+      ngx.ctx.a8_retries = selected_backend.retries
+   end
 
-   if upstream.host then
+   if selected_instances[1].host then -- FIXME
```
ijsnellf: What is the `-- FIXME` in reference to?

frankbu (approved):
LGTM

sidecar/nginx/lua/amalgam8.lua
rshriram: If the target instance requires the HTTP host header to be present and each target instance has a different HTTP host header (e.g., 3wesdf.foo.mybluemix.net, s23sdf.foo.mybluemix.net, etc.), the load balancing won't work. This becomes a problem only when you register multiple `hostnames` under the same service name, instead of IP addresses. I am looking at ways to address this. Right now, the only use case affected by this change will be those who are trying to map external services to multiple backends, such as "cloudant" to "1.mydb.cloudant.com", "2.mydb.cloudant.com".

ijsnellf (approved):

PR #371, opened 2016-11-01
updated examples to use new supervisor code

GregHanson:
This CANNOT be merged until there is a new sidecar released, i.e. 0.4.0->0.4.1, in which case WGET commands in all the dockerfiles in examples/ will need to be update to this version


examples/apps/bookinfo/reviews/reviews-wlpcfg/config.yaml
```diff
@@ -0,0 +1,24 @@
+register: true
+
```
rshriram, line 2: need 
`proxy: true`

examples/apps/helloworld/Dockerfile.sidecar
```diff
@@ -14,11 +14,14 @@
 
 FROM python:2-onbuild
 
-RUN mkdir -p /opt/microservices
+COPY filebeat.yml /etc/filebeat/filebeat.yml
```
rshriram, line 17: why do we need filebeat for helloworld?

examples/apps/helloworld/filebeat.yml
```diff
@@ -0,0 +1,422 @@
...
+  # Enable debug output for selected components. To enable all selectors use ["*"]
+  # Other available selectors are beat, publish, service
+  # Multiple selectors can be chained.
+  #selectors: [ ]
+
+  # Sets log level. The default log level is error.
+  # Available log levels are: critical, error, warning, info, debug
+  level: info
+
+
```
rshriram: file not needed

examples/apps/helloworld/run_filebeat.sh
```diff
@@ -0,0 +1,11 @@
...
+LOGSTASH=`echo $A8_LOGSTASH_SERVER`
+
+if [ -z "$LOGSTASH" ]; then
+  echo "Env var A8_LOGSTASH_SERVER is not set."
+  exit 1
+fi
+
+sed -e s/LOGSTASH_REPLACEME/\"$LOGSTASH\"/ /etc/filebeat/filebeat.yml >/tmp/filebeat.yml
+cp /tmp/filebeat.yml /etc/
+exec filebeat -c /etc/filebeat.yml
```
rshriram: file not needed

examples/docker-bookinfo.yaml
```diff
@@ -30,6 +30,8 @@ gateway:
     - controller
     - registry
   container_name: gateway
+  command: /usr/bin/run_filebeat.sh
+
```
rshriram, line 34: Why does the gateway have to run the run_filebeat.sh? Isn't it the stock sidecar?

examples/docker-bookinfo.yaml
```diff
@@ -56,28 +58,19 @@ ratings-v1:
 reviews-v1:
   image: amalgam8/a8-examples-bookinfo-reviews-sidecar:v1
   environment:
-    - A8_REGISTRY_URL=http://registry:8080
     - A8_SERVICE=reviews:v1
```
rshriram, line 60: Why do we need this environment variable?
IF its for nginx, then, we are already constructing it in sidecar. I am not sure if the reviews Java code is reading the A8_SERVICE environment variable.

examples/docker-bookinfo.yaml
```diff
@@ -56,28 +58,19 @@ ratings-v1:
...
-    - A8_LOG=true
-    - A8_REGISTER=true
     - A8_PROXY=true
-    - A8_REGISTRY_URL=http://registry:8080
     - A8_SERVICE=reviews:v2
-    - A8_ENDPOINT_PORT=9080
     - A8_LOGSTASH_SERVER='logstash:8092'
     - A8_CONTROLLER_URL=http://controller:8080
     - A8_CONTROLLER_POLL=5s
     - A8_REGISTRY_POLL=5s
```
rshriram, line 72: Can we move all this into config.yaml for reviews service? May be we can just have two configs. config-reviews-v1.yaml and config-reviews-v2v3.yaml.

This makes the deployment script look much more simple.

rshriram: except for A8_LOGSTASH_SERVER

examples/docker-helloworld.yaml
```diff
@@ -30,6 +30,7 @@ gateway:
     - controller
     - registry
   container_name: gateway
+  command: /usr/bin/run_filebeat.sh
```
rshriram, line 33: Once again, why the run_filebeat?

rshriram: Also, for helloworld, we need to use the env section and define the A8_SERVICE variable as the python code depends on it.

examples/k8s-bookinfo.yaml
```diff
@@ -50,11 +50,11 @@ spec:
         imagePullPolicy: IfNotPresent
         ports:
         - containerPort: 6379
+        args:
+        - /usr/bin/run_filebeat.sh
```
rshriram, line 54: same bug

rshriram (changes requested):

examples/docker-helloworld.yaml
GregHanson: this line is always needed - because we do not run filebeat by default.  The ENTRYPOINT for `sidecar:alpine` image is :
`ENTRYPOINT ["/usr/bin/a8sidecar"]`  
I thought the decision was not to run filebeat by default

examples/k8s-bookinfo.yaml
GregHanson: see comment above

examples/docker-helloworld.yaml
GregHanson: I didn't remove `A8_SERVICE` in my commits - was this broken to begin with?  Is there any CI testing for the stuff in `examples/`

examples/docker-bookinfo.yaml
GregHanson: doesn't it need this env var for registering with registry?

examples/docker-bookinfo.yaml
GregHanson: If you want to have multiple config files for a sidecar to choose from, we would need to add in some functionality for dynamically selecting which one to read on start up - which we currently cannot do because the order for loading NGINX config is:

```
1. read config values from provided yaml file if set
2. set config values from provided command line flags
3. overwrite anything else with any set environment variable
```

examples/apps/bookinfo/reviews/reviews-wlpcfg/config.yaml
GregHanson: can't be `proxy: true` because reviews:v1 does not proxy.  The `proxy` flag is set as an env var in the docker-compose yaml for reviews v2,v3

examples/docker-bookinfo.yaml
rshriram: but the sidecar is the one that is registering it with the registry, right? So, if that variable is missing, it should be able to construct it from the register section in config file.

examples/docker-bookinfo.yaml
rshriram: What I meant is, in the reviews folder, we can have two sets of configs.. config-reviews-noproxy.yaml and config-reviews-sidecar-proxy.yaml. Then in the build-services.sh script, we can create a temporary version config.yaml (copied from config-reviews-noproxy.yaml) for reviews-v1, or another version for reviews-v2/v3.

examples/docker-helloworld.yaml
rshriram: Sorry for the confusion. we need A8_SERVICE env var for the helloworld python containers. May be that can go as part of the config.yaml for helloworld, illustrating how to set env vars in the command section.

For the gateway, the run_filebeat should be part of its config.yaml right?

rshriram (approved):

PR #374, opened 2016-11-02
Sidecar and controller default HTTP client timeouts

ijsnellf:
Modifies the HTTP clients in the controller and sidecar to have default timeouts for the reasons described here: https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779#.etxsy5jbe

I was going to modify registry as well, but I'm not sure what the correct defaults would be.


rshriram:
So there are no specific defaults for registry/controller. Its at the TCP socket level. We can keep it consistent across the subsystems.. I suggest we set the same for registry client in sidecar as well.

zcahana:
The registry already sets (varying) timeouts for the HTTP clients used by its [registry client](https://github.com/amalgam8/amalgam8/blob/master/registry/client/client.go#L107), [Eureka client](https://github.com/amalgam8/amalgam8/blob/master/registry/store/eureka/client.go#L68) and the [Kubernetes client](https://github.com/amalgam8/amalgam8/blob/master/registry/store/kubernetes/client.go#L69).

The registry's replication/sync clients do not (must not) set timeouts as they use long running SSE streams.

Other than that, it seems that all http client used across the system now have a timeout set. Hooray!

zcahana (approved):

PR #377, opened 2016-11-03
Supervisor zombie cleanup

GregHanson:
handles grandchild process cleanup by listening for SIGCHLD and calling syscall.wait4().  also adds the new yaml config command option for handling terminating any daemon processes the command may start up


sidecar/supervisor/appsupervisor.go
```diff
@@ -23,6 +23,9 @@ import (
 
 	"time"
 
+	"sync"
+
+	"github.com/Shopify/toxiproxy/Godeps/_workspace/src/github.com/Sirupsen/logrus"
```
ijsnellf: This dependency needs to be fixed. It was probably imported because `logrus` is renamed just `log` in this file.

sidecar/supervisor/appsupervisor.go
```diff
@@ -173,7 +195,7 @@ func terminateSubprocesses(procs []*process, sig os.Signal) {
 		if time.Now().After(timeout) {
 			for _, proc := range procs {
 				if proc.Cmd.Process != nil {
-					proc.Cmd.Process.Kill()
+					syscall.Kill(-proc.Cmd.Process.Pid, syscall.SIGKILL)
```
ijsnellf: Two things
1) Do we always want to kill the group in this case, regardless of the KillGroup setting?
2) Is `syscall.Kill(...)` synchronous? If so, could we speed this up with a wait group?

sidecar/supervisor/appsupervisor.go
```diff
@@ -23,20 +23,25 @@ import (
 
 	"time"
 
+	"sync"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/amalgam8/amalgam8/sidecar/config"
 	"github.com/amalgam8/amalgam8/sidecar/register"
+	"github.com/Shopify/toxiproxy/Godeps/_workspace/src/github.com/Sirupsen/logrus"
```
rshriram: Why do we need this import ?

rshriram (commented):
Is it possible to have a test for this feature? Not sure how feasible this is in ginkgo.

sidecar/supervisor/appsupervisor.go
ijsnellf: It is an accidental auto-import.

ijsnellf:
TL;DR - I'm closing the PR because the logic is not necessary.

There is a lot of misinformation about zombie processes floating around, so I read the `wait` man pages and performed some experiments to confirm I understood the behavior correctly. According to the `wait` man pages:

> A child that terminates, but has not been waited for becomes a "zombie". The kernel maintains a minimal set of information about the zombie process (PID, termination status, resource usage information) in order to allow the parent to later perform a wait to obtain information about the child. As long as a zombie is not removed from the system via a wait, it will consume a slot in the kernel process table, and if this table fills, it will not be possible to create further processes. **If a parent process terminates, then its "zombie" children (if any) are adopted by init(1), which automatically performs a wait to remove the zombies.**

https://www.mankier.com/2/wait#Notes

Zombie processes can only exist while their parent process exists. They are adopted by `init(1)` when their parent dies and `init(1)` reaps them when they exit. _This includes processes with a grandparent._ I confirmed this behavior via experimentation on Ubuntu 16.04 LTS. 

Since grandchildren are adopted and reaped by `init(1)`, there is no need for us to reap grandchildren in our supervision. In fact in our tests we confirmed that we only receive `SIGCHLD` notifications for our immediate children. Grandchildren do not generate `SIGCHLD` notifications inside the `sidecar`, so we have no way of knowing the status of our grandchildren has even changed.

Since grandchildren are reaped by `init(1)` automatically on parent death there is no need to add extra logic to reap them in our supervision code and I am closing this PR.

zcahana:
@ijsnellf your observation is correct, but is missing a key point: the main process (entrypoint) in a docker container always runs as PID 1, assuming the responsibility to reap zombies. This is aptly explained in [Docker and the PID 1 zombie reaping problem](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/).

Here's my own experimentation, running _within_ a container, showing how a ill-behaving PID 1 adopts its orphan grandchild, after its original parent had died.

``` bash
zvic@zvic-vbox:~$ docker run -d --name bad_pid1 ubuntu:14.04 bash -c "bash -c 'sleep 20s & exec sleep 10s' & exec sleep 1h"
e84036fc2ade343c8e82e7d0a6b93d40ac96ee9f4a5101f1d2f4ddc5fb831043
zvic@zvic-vbox:~$ docker exec -it bad_pid1 bash
root@e84036fc2ade:/# ps -fA
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  1 09:25 ?        00:00:00 sleep 1h
root         8     1  0 09:25 ?        00:00:00 sleep 10s
root         9     8  0 09:25 ?        00:00:00 sleep 20s
root        10     0  1 09:25 ?        00:00:00 bash
root        27    10  0 09:25 ?        00:00:00 ps -fA
root@e84036fc2ade:/# ps -fA
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 09:25 ?        00:00:00 sleep 1h
root         8     1  0 09:25 ?        00:00:00 [sleep] <defunct>
root         9     1  0 09:25 ?        00:00:00 sleep 20s
root        10     0  0 09:25 ?        00:00:00 bash
root        28    10  0 09:25 ?        00:00:00 ps -fA
root@e84036fc2ade:/# ps -fA
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 09:25 ?        00:00:00 sleep 1h
root         8     1  0 09:25 ?        00:00:00 [sleep] <defunct>
root         9     1  0 09:25 ?        00:00:00 [sleep] <defunct>
root        10     0  0 09:25 ?        00:00:00 bash
root        29    10  0 09:25 ?        00:00:00 ps -fA
```

And here's another version of this experiment, where now PID 1 does reap its direct children (more aligned with what our sidecar does):

``` bash
zvic@zvic-vbox:~$ docker run -d --name bad_pid1 ubuntu:14.04 bash -c "bash -c 'sleep 20s & exec sleep 10s' & wait && exec sleep 1h"
e9f5980b2325785d336969e86f95ae129bc049ce0a2be90abc45ba69d81e6116
zvic@zvic-vbox:~$ docker exec -it bad_pid1 bash
root@e9f5980b2325:/# ps -fA
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  1 09:32 ?        00:00:00 bash -c bash -c 'sleep 20s & exec sleep 10s' & wait && exec sleep 1h
root         9     1  0 09:32 ?        00:00:00 sleep 10s
root        10     9  0 09:32 ?        00:00:00 sleep 20s
root        11     0  2 09:32 ?        00:00:00 bash
root        29    11  0 09:32 ?        00:00:00 ps -fA
root@e9f5980b2325:/# ps -fA
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 09:32 ?        00:00:00 sleep 1h
root        10     1  0 09:32 ?        00:00:00 sleep 20s
root        11     0  0 09:32 ?        00:00:00 bash
root        30    11  0 09:32 ?        00:00:00 ps -fA
root@e9f5980b2325:/# ps -fA
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 09:32 ?        00:00:00 sleep 1h
root        10     1  0 09:32 ?        00:00:00 [sleep] <defunct>
root        11     0  0 09:32 ?        00:00:00 bash
root        31    11  0 09:32 ?        00:00:00 ps -fA
```

sidecar/supervisor/appsupervisor.go
```diff
@@ -204,6 +198,8 @@ func terminateSubprocesses(procs []*process, sig os.Signal) {
 		time.Sleep(100 * time.Millisecond)
 	}
 
+	time.Sleep(100 * time.Millisecond)
```
rshriram: What is this for? A one liner comment would be helpful.

GregHanson: Just giving any processes who were forcibly killed the chance to exit - modified the placement of this sleep to be in the same block as the timeout logic and added some comments

sidecar/supervisor/appsupervisor.go
```diff
@@ -94,6 +99,9 @@ func (a *AppSupervisor) DoAppSupervision() {
 		}
 	}
 
+	// listen for SIGTERMS from any of the supervised processes and any orphans spun off from those
```
zcahana: Comment fix: we're listening for SIGCHLD, not SIGTERM.

sidecar/supervisor/appsupervisor.go
```diff
@@ -170,16 +186,43 @@ func terminateSubprocesses(procs []*process, sig os.Signal) {
...
+		// If services have not exited gracefully within the alotted time, force kill them all
 		if time.Now().After(timeout) {
 			for _, proc := range procs {
 				if proc.Cmd.Process != nil {
-					proc.Cmd.Process.Kill()
+					syscall.Kill(-proc.Cmd.Process.Pid, syscall.SIGKILL)
 				}
 			}
+			// Allow for processes to handle signal
+			time.Sleep(100 * time.Millisecond)
```
zcahana: Child processes _cannot_ handle SIGKILL, as it cannot be trapped. For my understanding, the OS kills them immediately, so this sleep isn't necessary.

zcahana (approved):
LGTM

PR #379, opened 2016-11-04
Removed health checks from demo apps

ijsnellf:
Reduces the amount of time it takes for services to show up when running through demo steps.


GregHanson (approved):
LGTM

PR #383, opened 2016-11-08
 Test updates to fix issues on OS X.

craigyam:


glide.yaml
```diff
@@ -65,4 +65,19 @@ import:
...
+  version: 2.1.0
+  subpackages:
+  - load
+
+- package: github.com/davecgh/go-spew
+  version: 2df174808ee097f90d259e432cc04442cf60be21
+  subpackages:
+  - spew
+
+- package: golang.org/x/sys
```
zcahana: Why do we need to vendor `golang.org/x/sys`? It doesn't seem to be referenced anywhere in the code (neither within `github.com/shirou/gopsutil`)...

glide.yaml
```diff
@@ -65,4 +65,19 @@ import:
...
-  version: db96a2b759cdef4f11a34506a42eb8d1290c598e
\ No newline at end of file
+  version: db96a2b759cdef4f11a34506a42eb8d1290c598e
+
+- package: github.com/shirou/gopsutil
+  version: 2.1.0
+  subpackages:
+  - load
+
+- package: github.com/davecgh/go-spew
```
zcahana: Why do we need to lock down a specific version of `github.com/davecgh/go-spew`? We only use it indirectly via `github.com/stretchr/testify`, which already vendors it.

glide.yaml
ericvn: Good question. One answer is if one needs to clear open source for use (which usually includes dependents), having specific versions may make that task easier. Non-specification of versions, means versions may get updated more frequently which requires clearance tasks needing to be done more frequently. I think that since we collapse all the dependent vendor directories into one, we can end up with something different than what testify includes. One comment from a glide issue on flattening the tree:

First dependency to declare that it needs a Ref wins
Any subsequent dependency that declares the same package, but a different Ref, will generate a warning
If you want to explicitly control, you should declare the package in the top glide.yaml

zcahana: @ericvn two comments on that:
1. vendor collapsing doesn't seem to work. There are both [`/vendor/github.com/davecgh/go-spew/spew`](https://github.com/craigyam/amalgam8/tree/osx_build_updates/vendor/github.com/davecgh/go-spew/spew) and [`vendor/github.com/stretchr/testify/vendor/github.com/davecgh/go-spew/spew`](https://github.com/craigyam/amalgam8/tree/osx_build_updates/vendor/github.com/stretchr/testify/vendor/github.com/davecgh/go-spew/spew) (and each is a different version of `go-spew`!). I believe the more-inner ones wins, import-wise.
2. Why stop at `go-spew`? There are several other transitive dependencies - `go-difflib`, `objx`, etc... (by the way, all come from `testify`).

craigyam: @zcahana You're right about golang.org/x/sys.  Not sure why the update pulled it in (I guess I must have cleaned up some code after the update that removed the dependency), but I ran it again and it gets removed.  I'll clean that up.

craigyam: @zcahana Any ideas why the A8 Makefile doesn't use the glide update --strip-vendor option?  That option is what collapses the vendor directory.  I see that SD registry uses, but A8 does not.  Was that intentionally removed?  Should we consider adding that back in?  I can clean up the glide.yaml to remove go-spew and sys dependencies.  Do we want to add the --strip-vendor option in this commit as well or have a separate PR?

ericvn: @zcahana I had thought Amalgam8 in it's Makefile told Glide to collapse the vendor directories. That isn't being done which is causing the multiple copies/versions. I think we want to flatten out the vendor directories, and remove the go-spew version from the Amalgam8 yaml and let the Godep (which is what testify uses) version be installed into the vendor directory (and it's version would end up in the Glide lock file).

Removing the go-spew version in the Amalgam8 glide.yaml resolves you original comment. Correct? The additional work would be adding the change to flatten the vendor directory.

zcahana: @ericvn @craigyam I too thought we were `--strip-vendor`'ing, but it must have got lost when we unified the Makefiles under the monorepo. Let's use this PR to add it back, as well as to unglide go-spew and x/sys.

zcahana (approved):

PR #396, opened 2016-11-10
New a8cli

gicaceres:
New a8cli


rshriram:
This needs to go into a separate branch.
Please rename to a8ctl
No binaries or tarballs.
Needs Makefile integration

zcahana (changes requested):
1. The CLI package should integrate with and follow the conventions of the sorrunding repo, rather than appear like a repo on its own. This means use the .gitignore, glide files, Makefile. 
2. Please reuse existing packages wherever possible: Registry client, Controller client, i18n, 
3. Doc needs to go into the [amalgam8.github.io](https://github.com/amalgam8/amalgam8.github.io) repo. We don't maintain other README's in here.
4. CLI output should probably avoid pretty-table formatting, to allow for easier scripting with it. Thoughts?
5. Test-only resources (i.e., `fake_rule.json`, `fake_rule.yaml`) shouldn't mix with source code. By convention, these should go into a `testdata` folder (which is also ignore by the `go` tool).
6. Should add a corresponding `cmd/a8ctl` package from which we'll build the binary

gicaceres:
We need to make multiple changes before this can be merged.

gicaceres:
Changes:
- New branch go_cli
- Renamed binary to a8ctl-beta
- Removed tarballs
- Makefile integration
- Registry client, Controller client can be updated later while we improve the CLI
- README file is necessary until the amalgam8.github.io is updated
- Moved test files into a testdata folder

Issue #173, opened 2016-09-08
Kubernetes adapter: import pod labels as instances tags


a8bot:
<a href="https://github.com/zcahana"><img src="https://avatars.githubusercontent.com/u/17428696?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zcahana](https://github.com/zcahana)**
_Wednesday Aug 17, 2016 at 10:27 GMT_
_Originally opened as https://github.com/amalgam8/registry/issues/44_

---

The current implementation parses the pod name to extract a version tag. This assumes the pod name is of the format `<rc base name>-<version>-<generated id>`, which is automatically generated for pods created from a replication controller named `<rc base name>-<version>`.

This has the benefit of doing a single query (or watch) to the `/endpoints` API endpoint, to get the list service names + endpoints + version tags. The downside is that it's based on loose conventions, doesn't support multiple tags, and doesn't integrate well with existing systems that already has a different name schemes or tag requirements.

What we'd like to do is to import labels attached to a pod and set them as registry instance tags.
This can be done using a single API call per-pod, i.e., whenever a service endpoints is added, make an API call to k8s to fetch the tags for that pod. Alternatively, we can place a watch on the `/pods` API endpoint, and cache their tags.

We also need to have the sidecar automatically fetch its own pod labels (--> tags), so that users will not have to specify the labels twice (in their k8s yaml's AND a8-sidecar configuration).
#### Open items:
1. What pod labels to import?
   - All `"key": "value"` pairs. This means that the registry instance tags will be formatted as "key=value", and so tags in routing/proxy rules will be specified so as well.
   - Only a label named `tags`, in which the value is a comma-separated list of tags, e.g.:  
     `"tags": "foo,bar,baz"`
2. Should we additionally import labels from related K8s objects, e.g. the replication controller that created the pod?

zcahana:
Resolved by #419.

Issue #172, opened 2016-09-08
Kubernetes adapter: re-implement using watches instead of polling


a8bot:
<a href="https://github.com/zcahana"><img src="https://avatars.githubusercontent.com/u/17428696?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [zcahana](https://github.com/zcahana)**
_Wednesday Aug 17, 2016 at 09:59 GMT_
_Originally opened as https://github.com/amalgam8/registry/issues/43_

---

See `GET /api/v1/namespaces/{namespace}/endpoints` on the [Kubernetes API documentation](http://kubernetes.io/docs/api-reference/v1/operations/).

This can reduce the number of API calls done, as well as accelerate the propagation of endpoint changes.

zcahana:
Resolved by #419.

PR #397, opened 2016-11-13
Pluggable service discovery backend for the sidecar

zcahana:
Enable the sidecar to have a pluggable service discovery backend to pull service instances from.
Currently supported backends are the Amalgam8 registry, Kubernetes and Eureka. Consul support to be added.

Key changes:
- Define `api.ServiceRegistry` and `api.ServiceDiscovery` interfaces, and provide implementations for a8/k8s/eureka backends (based on existing clients/catalog adapters).
- Modify the sidecar configuration and logic accodingly

This PR contributes to #348 (but doesn't resolve it - still needs doc'ing, examples)


zcahana:
The new registry configuration format supported by the sidecar:

``` yaml
registry:
    backend: kubernetes
    kubernetes:
        url: http://172.17.0.11:8080
        token: 12345
        namespace: mynamespace
```

or:

``` yaml
registry:
    backend: eureka
    eureka:
        urls:
         - http://172.17.0.11:8080
         - http://172.17.0.12:8080
         - http://172.17.0.13:8080
```

or

``` yaml
registry:
    backend: amalgam8
    amalgam8:
        url: http://172.17.0.11:8080
        token: 12345
        poll: 15s
```

The old registry configurartion format will be deprecated, but still supported (and is equivalent to the prev one):

``` yaml
registry:
    url: http://172.17.0.11:8080
    token: 12345
    poll: 15s
```

Also, added flags/environment variables corresponding to the above:
- `--registry_backend`, `A8_REGISTRY_BACKEND`
- `--kubernetes_url`, `A8_KUBERNETES_URL`
- `--kubernetes_token`, `A8_KUBERNETES_TOKEN`
- `--kubernetes_namespace`, `A8_KUBERNETES_NAMESPACE`
- `--eureka_url`, `A8_EUREKA_URL`

rshriram:
In kubernetes, why do we need to provide the API servers details? Won't it be available as part of the pod's config? The google docs seemed to indicate so. 

When we are using the kubernetes registry from sidecar, let's use the technique suggested by the google folks.

zcahana:
This PR only addresses having a pluggable discovery backend in the sidecar. It builds on the existing code for kubernetes/eureka adapters and a8 registry client, and refactors them into a common interface that can be used from the sidecar.

Yes, there's lots of room for improvement in the kubernetes adapter - including automatic configuration of API server credentials, switch to use official golang client, switch to use watch API, import pod labels as tags, ... I think these will be worked on as part of #172 and #173.

rshriram (approved):

PR #399, opened 2016-11-16
filebeat bug fix to make gremlin work

rshriram:
Fixes issue that caused filebeat to not start which subsequently caused Gremlin tests to fail.


ijsnellf (approved):
LGTM.

Issue #402, opened 2016-11-17
Can't extend a8sidecar nginx.conf with stream socket context by adding a file to /etc/nginx/conf.d


greglanthier:
I would like to use Amalgam8 for an application that is composed of web services as well as JMS, MongoDB, and JDBC servers.

IIUC it is not currently possible to proxy TCP or UDP sockets via Amalgam8 sidecars - presumably due to issues alluded to by [issues/4#issuecomment-245413281](https://github.com/amalgam8/amalgam8/issues/4#issuecomment-245413281).  I don't know if the Lua magic used by the a8sidecar can generate Nginx configuration for TCP and UDP sockets.

Theoretically we could extend the Nginx configuration used by the amalgam8/a8-sidecar:alpine Docker container to proxy TCP or UDP sockets on our own until the Lua thing is addressed, but when I tried I ran into the problems described in this [comment](https://github.com/amalgam8/amalgam8/issues/4#issuecomment-261077407).

> I have a ```mongo-proxy.conf``` file that looks like this:
>```
> stream {
>  upstream backend {
>    server <ip>:27017;
>  }
>  server {
>    listen 21212;
>    proxy_connect_timeout 1s;
>    proxy_timeout 1m;
>    proxy_pass backend;
>  }
>}
>```
>that I mount into `amalgam8/a8-sidecar:alpine` as `/etc/nginx/conf.d/mongo-proxy.conf` but at startup this fails since it looks like the default nginx.conf file loads `/etc/nginx/conf.d/*.conf` files into the http context.
>
>For example:
>
>mongo_sidecar    | time="2016-11-16T21:25:46Z" level=info msg="Starting NGINX" 
>mongo_sidecar    | time="2016-11-16T21:25:46Z" level=error msg="Failed to start NGINX service" error="exit status 1: nginx: [emerg] \"stream\" directive is not allowed here in /etc/nginx/conf.d/mongo_proxy.conf:1\n" 
>mongo_sidecar    | time="2016-11-16T21:25:46Z" level=warning msg="Registry listener f

One way to allow us to customise the Nginx configuration in the a8sidecar would be to do something like this in the [nginx.conf](https://github.com/amalgam8/amalgam8/blob/master/sidecar/nginx/conf/nginx.conf#L16-L33) configuration file:

```
http {
  ...
  include /etc/nginx/conf.d/*.http.conf;
  ...
}

stream {
  include /etc/nginx/conf.d/*.stream.conf;
}
```

[...]

PR #403, opened 2016-11-17
add stream block to nginx.conf

rshriram:
This PR addresses issue #402 reported by @greglanthier. Users can include configurations for proxying TCP services as part of their sidecar customization. This is an interim fix until we support generating TCP config files as part of the config reload process.

### DIFF
--- a/sidecar/nginx/conf/nginx.conf
+++ b/sidecar/nginx/conf/nginx.conf
@@ -33,5 +33,5 @@ http {
 }
 
 stream {
- include /etc/nginx/conf.d/*.stream.conf;
+ include /etc/nginx/conf.d/stream/*.conf;
 }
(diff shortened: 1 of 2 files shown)


sidecar/nginx/conf/nginx.conf
```diff
@@ -26,12 +26,12 @@ http {
   keepalive_requests 10000;
   types_hash_max_size 2048;
 
-  include /etc/nginx/conf.d/*.http.conf;
+  include /etc/nginx/conf.d/*.conf;
```
frankbu, line 29: won't this also pull in the xxx.stream.conf files?

rshriram: ugh.. Good point. this is the default nginx config.. But we also need to be able to let people pull in stream config stuff. May be we put the stream files under conf.d/stream/*.conf ?

rshriram:
fixes #402

frankbu (approved):
LGTM

Issue #401, opened 2016-11-17
Zombie reaping logic hijacks child exit status


zcahana:
The newly added zombie reaping logic reacts to SIGCHLD by doing a wait() syscall.
This hijacks the child's exit status from competing wait() calls, interfering with......

1. Nginx updater code which [fails to verify that nginx is running](https://github.com/amalgam8/amalgam8/blob/master/sidecar/proxy/nginx/service.go#L82), so doesn't update nginx with new rules/services:
```
time="2016-11-17T12:52:44Z" level=error msg="Could not get status of NGINX service" error="waitid: no child processes" 
time="2016-11-17T12:52:44Z" level=warning msg="Controller listener failed" error="waitid: no child processes" 
```

2. Command-type healthchecks (e.g., `file:///bin/ls` for the productpage microservice) which [fails to read the command exit status](https://github.com/amalgam8/amalgam8/blob/master/sidecar/register/healthcheck/command.go#L87
), and triggers deregistration:
```
time="2016-11-17T12:46:04Z" level=info msg="Starting Amalgam8 service registration agent" service_name=productpage 
time="2016-11-17T12:46:04Z" level=info msg="Service successfully registered with Amalgam8" instance_id=c366a7b3223239bc service_name=productpage 
time="2016-11-17T12:48:29Z" level=info msg="Stopping Amalgam8 service registration agent" service_name=productpage 
time="2016-11-17T12:48:29Z" level=info msg="Attempting to deregister service with Amalgam8" instance_id=c366a7b3223239bc service_name=productpage 
time="2016-11-17T12:48:29Z" level=info msg="Service successfully deregistered with Amalgam8" instance_id=c366a7b3223239bc service_name=productpage 
time="2016-11-17T12:48:59Z" level=info msg="Starting Amalgam8 service registration agent" service_name=productpage 
time="2016-11-17T12:48:59Z" level=info msg="Service successfully registered with Amalgam8" instance_id=c366a7b3223239bc service_name=productpage 
time="2016-11-17T12:49:29Z" level=info msg="Stopping Amalgam8 service registration agent" service_name=productpage 
time="2016-11-17T12:49:29Z" level=info msg="Attempting to deregister service with Amalgam8" instance_id=c366a7b3223239bc service_name=productpage 
time="2016-11-17T12:49:29Z" level=info msg="Service successfully deregistered with Amalgam8" instance_id=c366a7b3223239bc service_name=productpage

rshriram:
[...]

PR #407, opened 2016-11-17
removed overzealous zombie process clean up and switched to dumb-init

GregHanson:
the current zombie process cleanup was receiving SIGCHLD's for commands started in other parts of the sidecar code (see issue #401 ).  Switched over to use dumb-init as PID to handle zombie process cleanup

### DIFF
--- a/docker/Dockerfile.sidecar.alpine
+++ b/docker/Dockerfile.sidecar.alpine
@@ -12,102 +12,6 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-# FROM alpine:latest
-
-# RUN mkdir /var/log/nginx
-
-# # Docker Build Arguments
-# ARG RESTY_VERSION="1.9.15.1"
-# ARG RESTY_OPENSSL_VERSION="1.0.2h"
-# ARG RESTY_PCRE_VERSION="8.38"
-# ARG RESTY_J="1"
-# ARG RESTY_CONFIG_OPTIONS="\
-#     --with-file-aio \
-#     --with-http_addition_module \
-#     --with-http_auth_request_module \
-#     --with-http_dav_module \
-#     --with-http_flv_module \
-#     --with-http_geoip_module=dynamic \
-#     --with-http_gunzip_module \
-#     --with-http_gzip_static_module \
-#     --with-http_image_filter_module=dynamic \
-#     --with-http_mp4_module \
-#     --with-http_perl_module=dynamic \
-#     --with-http_random_index_module \
-#     --with-http_realip_module \
-#     --with-http_secure_link_module \
-#     --with-http_slice_module \
-#     --with-http_ssl_module \
-#     --with-http_stub_status_module \
-#     --with-http_sub_module \
-#     --with-http_v2_module \
-#     --with-http_xslt_module=dynamic \
-#     --with-ipv6 \
-#     --with-mail \
-#     --with-mail_ssl_module \
-#     --with-md5-asm \
-#     --with-pcre-jit \
-#     --with-sha1-asm \
-#     --with-stream \
-#     --with-stream_ssl_module \
-#     --with-threads \
-#     --conf-path=/etc/nginx/nginx.conf \
-#     --pid-path=/var/run/nginx.pid \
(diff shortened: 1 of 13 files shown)


ijsnellf:
We'll need to update the docs as well if we merge this change.

ijsnellf (approved):
LGTM, but it'd be ideal to get input from @zcahana before we merge it.

docker/Dockerfile.sidecar.alpine
```diff
@@ -132,6 +36,9 @@ RUN set -ex \
     && mkdir /lib64 \
     && ln -s /lib/libc.musl-x86_64.so.1 /lib64/ld-linux-x86-64.so.2
 
+RUN apk --no-cache add openssl
+RUN wget -O /usr/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.2.0_amd64
+RUN chmod +x /usr/bin/dumb-init
```
rshriram, line 42: is the openssl package needed ? openresty:alpine already comes with openSSL latest version

GregHanson: it apparently does not because the alpine image fails to build without adding openssl

zcahana: Can we do this in a single RUN directive so that we get one layer?

docker/Dockerfile.sidecar.ubuntu
```diff
@@ -32,13 +32,16 @@ RUN set -ex \
     \
     && rm -rf /tmp/filebeat*
 
+RUN curl -sSL -o /usr/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.2.0_amd64
+RUN chmod +x /usr/bin/dumb-init
```
zcahana: ditto

scripts/a8sidecar.sh
```diff
@@ -63,3 +63,10 @@ tar -xzf /tmp/a8sidecar-${A8SIDECAR_RELEASE}-linux-amd64.tar.gz -C /
 #Cleanup
 rm -rf ${A8TMP}
 rm /tmp/a8sidecar-${A8SIDECAR_RELEASE}-linux-amd64.tar.gz
+
+if [ $HAVE_WGET -eq 1 ]; then
```
zcahana, line 67: I suggest we not including dumb-init in our installation script, as it is _not_ a dependency for the amalgam8 sidecar.
The sidecar can perfectly operate without it, because it properly reaps its child processes, and so does nginx.

However, given that using a proper init process is considered best-practices when running multiple processes in a container, we should install and use it our prebuilt sidecar images, and probably add a doc recommendation so that users who build their own images will install dumb-init/runit/whatever, and refer them to our Dockerfile as an example.

testing/apps/details/Dockerfile.sidecar
```diff
@@ -19,4 +19,4 @@ COPY details /opt/microservices/
 
 EXPOSE 9080
 
-ENTRYPOINT ["a8sidecar", "/opt/microservices/details", "9080" ]
\ No newline at end of file
+ENTRYPOINT ["/usr/bin/dumb-init", "--", "a8sidecar", "/opt/microservices/details", "9080" ]
```
zcahana: This is not directly related to this PR, but the bookinfo/helloworld Dockerfiles shouldn't redefine the ENTRYPOINT, but rather inherit it from the base image, and add their arguments as CMD:

```
CMD [ "/opt/microservices/details", "9080" ]
```

zcahana (changes requested):
Overall looks ok, just a few little comments.

zcahana:
@GregHanson @ijsnellf are you guys still working this? We need to get issue #401 resolved as its breaks the `master` branch for now.

One suggestion was to keep the zombie reaping logic code around, but use it only if running as PID 1, and make sure to spawn a child process for the all other functions of the sidecar. This will also require adding code to forward any signals down to the child sidecar process. I myself am ok with both approaches (dumb-init or are own zombie reaper).

GregHanson:
Closing.  Resolved with pull request #421.

PR #409, opened 2016-11-18
Support discovery REST API in sidecar agent

craigyam:
Instead of the sidecar pushing new updates to the proxy (whether its
nginx or envoy), the sidecar could also listen on a localhost endpoint
where the proxy could poll periodically to obtain upstream instances
(the registry data). This model would decouple the proxy implementation
from the sidecar agent and allow us to switch between different proxies
as well.

At minimum, the sidecar needs to support Envoy's API endpoint for
service discovery.

### DIFF
--- a/sidecar/config/config.go
+++ b/sidecar/config/config.go
@@ -152,6 +152,8 @@ type Config struct {
 	Commands []Command `yaml:"commands"`
 
 	Debug string
+
+	DiscoveryPort int
 }
 
 // New creates a new Config object from the given commandline flags, environment variables, and configuration file context.
@@ -256,6 +258,7 @@ func (c *Config) loadFromContext(context *cli.Context) error {
 	loadFromContextIfSet(&c.Dnsconfig.Domain, dnsConfigDomainFlag)
 	loadFromContextIfSet(&c.LogLevel, logLevelFlag)
 	loadFromContextIfSet(&c.Debug, debugFlag)
+	loadFromContextIfSet(&c.DiscoveryPort, discoveryPortFlag)
 
 	if context.IsSet(serviceFlag) {
 		name, tags := parseServiceNameAndTags(context.String(serviceFlag))
--- a/sidecar/config/config.go
+++ b/sidecar/config/config.go
@@ -352,7 +352,8 @@ func (c *Config) Validate() error {
 	validators = append(validators,
 		IsInSet("Registry backend", c.Registry.Backend, []string{Amalgam8Backend, KubernetesBackend, EurekaBackend}),
 		IsEmptyOrValidURL("Amalgam8 Registry URL", c.Registry.Amalgam8.URL),
-		IsEmptyOrValidURL("Kubernetes URL", c.Registry.Kubernetes.URL))
+		IsEmptyOrValidURL("Kubernetes URL", c.Registry.Kubernetes.URL),
+		IsInRange("Service Discovery Port", c.DiscoveryPort, 1, 65535))
 	for _, url := range c.Registry.Eureka.URLs {
 		validators = append(validators, IsEmptyOrValidURL("Eureka URL", url))
 	}
--- a/sidecar/register/discovery/discovery.go
+++ b/sidecar/register/discovery/discovery.go
@@ -1,0 +1,174 @@
+// Copyright 2016 IBM Corporation
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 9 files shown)


sidecar/sidecar.go
```diff
@@ -101,6 +101,25 @@ func Run(conf config.Config) error {
 			logrus.WithError(err).Error("Could not create service discovery backend adapter")
 			return err
 		}
+
+		if conf.DiscoveryPort == 0 {
```
ijsnellf: Instead of setting the default here, can you remove this segment of code and instead put the default inside of `sidecar/config/default.go`? That way we support config files, env vars, and args.

sidecar/config/config.go
```diff
@@ -256,6 +258,7 @@ func (c *Config) loadFromContext(context *cli.Context) error {
 	loadFromContextIfSet(&c.Dnsconfig.Domain, dnsConfigDomainFlag)
 	loadFromContextIfSet(&c.LogLevel, logLevelFlag)
 	loadFromContextIfSet(&c.Debug, debugFlag)
+	loadFromContextIfSet(&c.DiscoveryPort, discoveryPortFlag)
```
ijsnellf, line 261: Could you add a validator for the discovery port to the `Validate` function? Something like `IsInRange("Service Discovery Port", c.DiscoveryPort, 1, 65535)`.

sidecar/register/discovery/discovery.go
```diff
@@ -0,0 +1,174 @@
...
+// Hosts is the array of hosts returned by the GET registration
+type Hosts struct {
+	Hosts []Host `json:"hosts"`
+}
+
+// Host is the endpoint and tag data for a service instance
+type Host struct {
+	IPAddr string            `json:"ip_address"`
+	Port   uint16            `json:"port"`
+	Tags   map[string]string `json:"tags"`
```
ijsnellf, line 47: Alternately, we know that Envoy only supports `az`, `canary` and `load_balance_weight` fields, so we can explicitly declare them here. Not necessary for this PR, though.

Reference:
https://lyft.github.io/envoy/docs/configuration/cluster_manager/sds_api.html#config-cluster-manager-sds-api

sidecar/register/discovery/discovery.go
```diff
@@ -0,0 +1,174 @@
...
+	resp := Hosts{
+		Hosts: translate(instances),
+	}
+
+	w.WriteHeader(http.StatusOK)
+	w.WriteJson(&resp)
+	return
+}
+
+func translate(instances []*api.ServiceInstance) []Host {
```
ijsnellf, line 93: This logic will be expanded. This looks good for a simple 1st pass.

sidecar/register/server.go
```diff
@@ -0,0 +1,136 @@
...
+
+// NewDiscoveryServer creates a new server based on the provided configuration options.
+// Returns a valid Server interface on success or an error on failure
+func NewDiscoveryServer(conf *Config) (Server, error) {
+	if conf == nil {
+		return nil, errors.New("null Discovery configuration provided")
+	}
+
+	s := &server{
+		config: &*conf,
```
ijsnellf: What does this do?

sidecar/register/server.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+	}()
+
+	return nil
+}
+
+func (s *server) Stop() {
+	s.logger.Info("Stopping rest server")
+
+	if err := s.listener.Close(); err != nil {
```
ijsnellf: Should we ensure `s.listener != nil`?

sidecar/register/server.go
```diff
@@ -0,0 +1,136 @@
...
+	restAPI := rest.NewApi()
+
+	restAPI.Use(
+		&rest.RecoverMiddleware{},
+		&rest.TimerMiddleware{},
+		&rest.RecorderMiddleware{},
+		&rest.ContentTypeCheckerMiddleware{},
+	)
+
+	log.SetOutput(s.logger.Logger.Out)
```
ijsnellf: Doesn't this override all log output except for `s.logger.Logger.Out` output? The `registry/util/logging` package should already be directing output correctly, I thought.

ijsnellf (changes requested):
Looks pretty good. Few comments / recommendations.

sidecar/config/config.go
craigyam: Done

sidecar/register/server.go
craigyam: Hm, not sure.  Took this from registry.  Seems to work fine without the &* so I took it out.

sidecar/register/server.go
craigyam: Done

sidecar/register/server.go
craigyam: Done

ijsnellf (approved):
LGTM, but let me create an envoy branch to push this to instead of master.

PR #416, opened 2016-11-21
Makefile update for grep on a Mac

ericvn:
The -x on the grep does not work the same on a Mac so we end up doing all the tasks against the vendor/* sub-directories as well which results in failures. Removing the -x solves the problem on the Mac and I tried against the Dev VM and saw no issues.


zcahana (approved):
I too see no issues with removing the `-x` flag. With our directory structure, this should work exactly the same.

PR #417, opened 2016-11-23
Update glide and vendor tree

zcahana:
- Bumped up glide version to v0.12.3 (sorted diff-friendly lock file, vendor flattening fixes, ...)
- Switched to semantic versions where applicable in glide.yaml
- Removed redundant 'subpackages' clauses from glide.yaml
- Removed redundant and deprecated --strip-vcs and --update-vendored flags from Makefile
- Moved test dependencies to the end of glide.yaml
- Regenerated the vendor tree from scratch


ijsnellf (approved):
LGTM

PR #419, opened 2016-11-29
Kubernetes discovery adapter rework

zcahana:
This PR reworks the existing Kubernetes adapter code, with the following key changes:

- The official [github.com/kubernetes/client-go](https://github.com/kubernetes/client-go) client package is now used, instead of a simple HTTP-based client. The official client includes support for watches, self-populating caches, API objects, etc.
- The adapter logic is now based on watches rather than polling (i.e., almost immediate response to changes)
- Instance tags are based on pod labels, instead of pod names conventions. Each label is mapped into a corresponding "key=value" tag. Dynamic label updates are watched and reflected as well. Additionally, pod annotations are mapped to instance metadata.
- The adapter can read URL/token configuration from the pod [service account](http://kubernetes.io/docs/user-guide/service-accounts/) (when running within a Kubernetes pod), instead of passing explicit configuration.
- The adapter can be configured to target a predefined namespace (defaults to "default" namespace).

Note: the `github.com/kubernetes/client-go` package(s), combined with its own dependencies, is very large (50+ MB). This also affect the compile time for the registry/sidecar (which use the adapter), as well as their produced binary sizes (~47MB). Pending PR #418 takes care of significantly shrinking the binaries.


zcahana:
Note to reviewers: the `registry/adapters/kubernetes/kubernetes.go` file is essentially rewritten from scratch, so I would suggest to review it as a whole, instead of in the "diff" view.

pkg/datastructures/stringset.go
```diff
@@ -0,0 +1,116 @@
...
+	for elem := range set {
+		if newSet == nil || !otherSet.Exists(elem) {
+			newSet.Add(elem)
+		}
+	}
+
+	return newSet
+}
+
+func min(x, y int) int {
```
rshriram: These functions seem arbitrary in this file. Why can't we use math.min and math.max ?

registry/adapters/kubernetes/kubernetes.go
```diff
@@ -16,73 +16,179 @@ package kubernetes
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 )
 
 const (
-	module = "K8SADAPTER"
+	// EndpointsCacheResyncPeriod is the period in which we do a full resync of the endpoints cache.
+	EndpointsCacheResyncPeriod = time.Duration(60) * time.Second
+
+	// PodCacheResyncPeriod is the period in which we do a full resync of the endpoints cache.
```
rshriram: what is the difference between these two? The comments seem to be the same for both.

registry/adapters/kubernetes/kubernetes.go
```diff
@@ -16,73 +16,179 @@ package kubernetes
...
+	podController cache.ControllerInterface
 
+	// services maps a service name to a list of service instances
 	services map[string][]*api.ServiceInstance
 
-	logger *log.Entry
-	sync.RWMutex
+	// servicePods maps a service name to a set of pods implementing it
+	servicePods map[string]datastructures.StringSet
+
```
rshriram: What is the difference between ServiceInstance and servicePods ?

registry/adapters/kubernetes/kubernetes.go
```diff
@@ -107,8 +213,8 @@ func (a *Adapter) ListInstances() ([]*api.ServiceInstance, error) {
 
 // ListServiceInstances queries for the list of service instances currently registered for the given service.
 func (a *Adapter) ListServiceInstances(serviceName string) ([]*api.ServiceInstance, error) {
```
rshriram: Once again, what is the difference between the two functions, ListInstances and ListServiceInstances ?

registry/adapters/kubernetes/kubernetes.go
```diff
@@ -117,84 +223,276 @@ func (a *Adapter) ListServiceInstances(serviceName string) ([]*api.ServiceInstan
...
+	for _, obj := range a.endpointsCache.List() {
+		endpoints := obj.(*v1.Endpoints)
+		if endpoints.Name == serviceName {
+			return endpoints
+		}
+	}
+	return nil
+}
+
+// getCachedServiceEndpoints returns the cached pod resource for the given UID, or nil if doesn't exist.
```
rshriram: nit. copy paste error

rshriram (changes requested):
Comments and questions inline. We also need to filter out what we cache locally, since we are looking at a lot of duplicated information across pods on the same host.

Updating the examples would also be helpful.
Simultaneously, it would be very very helpful if we could make the sidecar use this library to fetch K8S instances instead of querying the registry service.

pkg/datastructures/stringset.go
zcahana: math.Min/Max work with floats, so are inconvenient and inefficient for use with integers.
We can, however, move these to our own `amalgam8/pkg/math` package, to be reused across the project.

registry/adapters/kubernetes/kubernetes.go
zcahana: first one is used for the endpoints cache, and the second one for the pods cache. I'll fix the comments.

registry/adapters/kubernetes/kubernetes.go
zcahana: `api.ServiceInstance` is a8's representation of an instance of a service. It is built and stored in the `services` map, so that it won't have to be recalculated with every `ListInstances()` call.
`servicePods` maps service name to a set of pod UIDs that implement it. This is maintained, together with the opposite mapping (pod UID --> services), so that we can supporting dynamic labels update on pods.

registry/adapters/kubernetes/kubernetes.go
zcahana: One list all instances, and the other one list instances for a particular service.
This is part of the [`api.ServiceDiscovery`](https://github.com/amalgam8/amalgam8/blob/master/registry/api/types.go#L24) interface, so required to be here in order to implement it.

zcahana:
@rshriram I'm already working on migrating the (k8s) examples to use the adapter. This will be done on a separate PR. However, merging it will have to wait until the adapter rework is released, and the updated examples images are in Dockerhub.

Also, the sidecar can already use the adapter to talk to k8s directly. This was enabled in PR #397 a while ago.

registry/adapters/kubernetes/kubernetes.go
rshriram: Excellent. Can you just stick this into the comments please?

pkg/datastructures/stringset.go
rshriram: Is it more inconvenient than maintaining our own version of the math package for integers? The type casting inefficiency is not so high. This is micro optimization considering all the string operations we are doing. But I leave it to you.

registry/adapters/kubernetes/kubernetes.go
rshriram: Once again, stick this into comments please.

rshriram:
Yes I am aware of PR #397. However I was hoping a zero config example with this PR (i.e. read from service accounts). 

Let's get this PR fixed and merged. As you said, the examples would depend on this PR. 

On a side note: do we need explicit configuration for k8s URL at all ? According to google docs, the client was supposed to be able to "automatically" connect to the API server (could find out exactly where the magic of deciphering the k8s API server IP address  was happening). 

If no explicit configuration is needed, can we get rid of related code? 

Secondly, please update the docs as well. Else we would end up doing a major rework of docs for the full k8s feature set.

registry/adapters/kubernetes/kubernetes.go
zcahana: That's... already there :)

rshriram: Oh. nm my bad.

zcahana:
Zero* configuration is already supported in this PR: the adapter can automatically fetch the K8s URL/Token from the service account. It also supports explicit URL/Token configuration, so that it can run out of a K8s pod.

\* Zero is actually not zero, because you still need to configure *what* type of registry you're working with - be it A8 registry, Eureka or K8s. But with K8s, that's all you need to configure.

zcahana:
More updates underway, do not merge.

pkg/datastructures/stringset.go
frankbu: I think Zvi's approach of just making his own simple min/max functions is typical: http://mrekucci.blogspot.ca/2015/07/dont-abuse-mathmax-mathmin.html

zcahana:
Add 3 commits addressing code review comments from @zachidan and @rshriram:
- Move min/max functions to their own package
- Switch to use a lower-level RESTClient
- Bugfix in bookkeeping of the podServices data structure
- Index pods by their names rather than by UIDs
- Fetch cached pods/endpoints by key rather than by linear search
- Comments fix
- Use a regular Mutex instead of RWMutex (single reader expected)
- Use finer-grain locking
- Use a single-worker workqueue to process watch events
- Avoid re-processing endpoints/pod updates when resource version hasn't changed (on full resync)

Also rebased the branch on master (to get the fix for issue #401), which seems to mess a bit with t the commits order.

zcahana:
Two additional notes:
1. I've looked into filtering out pod data from our cache, so that we keep only what we need (names, labels and annotations). Apparently, when using the build-in k8s client cache, there's no easy way to do so, if at all. The alternative is probably to stop use that cache, and maintain our own. So, to keep things simple, let's live with the current approach for now.
2. The integration tests on Travis are failing because the registry, when linked with the k8s client packages,  doesn't produce a statically linked binary, which fails to start on a "scratch"-based container. PR #418 includes a fix to our compile/links flags, so that we get a proper static linked binary. Once this is merged, we can rebase this branch and get a green build.

rshriram (approved):
Looks good for the initial pass.

rshriram:
@zcahana the PR looks good for first pass. Seems the branch is out of sync (as per GitHub UI). Do whatever sync you need to and merge it.

rshriram:
fixes #172, fixes #173

PR #420, opened 2016-11-29
Add sidecar config for SSL to allow for secure sidecar communication 

paulben:
Allow ssl for: inbound to nginx, outbound from nginx to service, and healthchecks.


testing/apps/details/Dockerfile.sidecar
```diff
@@ -12,11 +12,11 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-FROM amalgam8/a8-sidecar:alpine
+FROM amalgam8/a8-sidecar:latest
```
rshriram: Why this change? Changes to this file seem unnecessary

testing/apps/productpage/Dockerfile.sidecar
```diff
@@ -12,11 +12,11 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-FROM amalgam8/a8-sidecar:alpine
+FROM amalgam8/a8-sidecar:latest
```
rshriram: same as above

testing/apps/ratings/Dockerfile.sidecar
```diff
@@ -12,11 +12,11 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-FROM amalgam8/a8-sidecar:alpine
+FROM amalgam8/a8-sidecar:latest
```
rshriram: same as above

testing/apps/reviews/Dockerfile.sidecar
```diff
@@ -12,7 +12,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-FROM amalgam8/a8-sidecar:alpine
+FROM amalgam8/a8-sidecar:latest
```
rshriram: same as above

sidecar/register/healthcheck/http.go
```diff
@@ -45,13 +47,33 @@ func NewHTTP(conf config.HealthCheck) (Check, error) {
...
+		client = &http.Client{Timeout: conf.Timeout,}	
+	}
 	return &HTTP{
 		url: conf.Value,
-		client: &http.Client{
-			Timeout: conf.Timeout,
-		},
-		method: conf.Method,
+		client: client,
+        method: conf.Method,
```
rshriram: `make precommit` should resolve these bad indents..

sidecar/config/flags.go
```diff
@@ -22,31 +22,36 @@ import (
...
-	healthchecksFlag        = "healthchecks"
-	logLevelFlag            = "log_level"
-	dnsFlag                 = "dns"
-	dnsConfigPortFlag       = "dns_port"
-	dnsConfigDomainFlag     = "dns_domain"
-	debugFlag               = "debug"
+	configFlag                  = "config"
+	registerFlag                = "register"
+	proxyFlag                   = "proxy"
+	proxyTypeFlag               = "proxy_type"
```
rshriram: may be rename to `proxy_tls_on` flag?

sidecar/config/default.go
```diff
@@ -22,10 +22,18 @@ var DefaultConfig = Config{
 	Proxy:    false,
 	DNS:      false,
 
+	ProxyConfig: ProxyConfig{
+		Type         : "http",
```
rshriram: May be rename to `TLS : false` ?

sidecar/config/default.go
```diff
@@ -68,6 +76,8 @@ var DefaultConfig = Config{
 
 	HealthChecks: nil,
 
+	HealthChecksCACertPath: "",
```
rshriram: Can be moved under HealthChecks config block.

sidecar/config/config.go
```diff
@@ -121,15 +124,25 @@ const (
...
+	Timeout    time.Duration `yaml:"timeout"`
+	Method     string        `yaml:"method"`
+	Code       int           `yaml:"code"`
+	Args       []string      `yaml:"args"`
+	CACertPath string        `yaml:"ca_cert_path"`
 }
 
+// Proxy configuration.
+type ProxyConfig struct {
+	Type         string `yaml:"type"`
```
rshriram: rename to 
```go
Tls bool `yaml: "tls_on"
``` 
or something

sidecar/config/config.go
```diff
@@ -146,6 +159,9 @@ type Config struct {
 	Supervise bool `yaml:"supervise"`
 
 	HealthChecks []HealthCheck `yaml:"healthchecks"`
+	HealthChecksCACertPath string `yaml:"healthchecks_ca_cert_path"`
```
rshriram: This should go into HealthChecks struct.

sidecar/config/config.go
```diff
@@ -426,3 +450,66 @@ func parseServiceNameAndTags(service string) (name string, tags []string) {
...
+	err = svcTemplate.Execute(file, a8ServicesData)
+	if err != nil {
+		return err
+	}
+	err = file.Close()
+	if err != nil {
+		return err
+	}
+	return err
+}
```
rshriram: This entire piece of code can go into a separate file proxyconfig.go or something, so that we can use it as a generic interface to generate configuration for any underlying proxy implementation (nginx/envoy/etc)

sidecar/config/config_test.go
```diff
@@ -360,12 +400,14 @@ log_level: debug
...
 			Expect(c.HealthChecks[0].Method).To(Equal("GET"))
 			Expect(c.HealthChecks[0].Code).To(Equal(200))
+			Expect(c.HealthChecks[0].CACertPath).To(Equal("/etc/certs/ca1.pem"))
 			Expect(c.HealthChecks[1].Type).To(Equal("http"))
 			Expect(c.HealthChecks[1].Value).To(Equal("http://localhost:8082/health2"))
 			Expect(c.HealthChecks[1].Interval).To(Equal(time.Duration(30) * time.Second))
 			Expect(c.HealthChecks[1].Timeout).To(Equal(time.Duration(3) * time.Second))
 			Expect(c.HealthChecks[1].Method).To(Equal("POST"))
 			Expect(c.HealthChecks[1].Code).To(Equal(201))
+			Expect(c.HealthChecks[1].CACertPath).To(Equal("/etc/certs/ca2.pem"))
```
rshriram, line 419: Thanks for adding the test cases for config. It would be nice to test the code in the healthcheck logic that instantiates a TLS client object for health checking.

rshriram (changes requested):
Thanks for this PR. My comments are inline.

testing/apps/details/Dockerfile.sidecar
paulben: Because otherwise the sidecar image doesn't pick up my go code changes, it pulls the alpine image.

sidecar/config/flags.go
paulben: The intent was to be consistent with endpoint:

```go
endpointTypeFlag = "endpoint_type"
```

If that consistency is not what we want then I can easily enough make it a bool flag.

sidecar/config/config.go
paulben: There is CACertPath in HealthChecks struct. This one is intended for ease of config via envir var and cmd line. The intent is to be similar to other health check stuff - you can specify the URL easily with A8_HEALTHCHECKS but for more detailed config have to use yaml. 

If I remove it we're basically saying health check tls must be configured in yaml...right?

paulben: Specifically HealthChecksCACertPath allowed me to test with this instead of yaml:

``` yaml
helloworld-v1:
  image: amalgam8/a8-examples-helloworld-sidecar:v1
  volumes:
   - 'certs:/opt/my/certs'
  environment:
    - A8_REGISTRY_URL=http://registry:8080
    - A8_SERVICE=helloworld:v1
    - A8_REGISTER=true
    - A8_ENDPOINT_PORT=5000
    - A8_ENDPOINT_TYPE=https
    - A8_HEALTHCHECKS=https://localhost:5000/health
    - A8_HEALTHCHECKS_CA_CERT_PATH=/opt/my/certs/ca_trust.pem
    - A8_LOG_LEVEL=debug
  external_links:
    - registry
```

sidecar/config/config.go
paulben: Will do.

testing/apps/details/Dockerfile.sidecar
paulben: I backed these changes out and tested again and these are not needed, I'll revert them in the next commit.

sidecar/config/flags.go
rshriram: So, we have to consider the case that the proxy could be just doing TCP proxying and no HTTP. In which case, TLS still applies. Which is why I suggested TLS being a boolean attribute.

sidecar/config/config.go
rshriram: I was really trying to cut down on the number of environment variables needed. We are trying to force people to using the YAML config for most part instead of creating umpteen number of CLI flags, as it tends to get more confusing.

In this case, I would say, yes. We are probably forcing people to use YAML config if they want to use HTTPs health checks reliably.

@zcahana @gicaceres @ijsnellf any thoughts?

sidecar/config/config.go
```diff
@@ -426,3 +450,66 @@ func parseServiceNameAndTags(service string) (name string, tags []string) {
 	}
 	return
 }
+
+type A8ServicesNginxConfData struct {
```
ijsnellf: Let's avoid prefixing structs/consts/variables with `A8` anywhere since this is the A8 code base. A8 is implicit.

This comment applies to the entire commit.

sidecar/config/config.go
```diff
@@ -426,3 +450,66 @@ func parseServiceNameAndTags(service string) (name string, tags []string) {
...
+// Generate proxy config
+func (c *Config) GenProxyConfig() error {
+    // Generate amalgam8-services.conf 
+    err := c.GenNginxA8ServicesConf()
+    return err
+}
+
+func (c *Config) GenNginxA8ServicesConf() error {
+	var a8ServicesData *A8ServicesNginxConfData
+	logrus.Debug("GenProxyConfig " + strconv.FormatBool(c.Proxy) + " " + c.ProxyConfig.Type)
```
ijsnellf: Alternate syntax that is a bit easier to read:
`logrus.Debugf("GenProxyConfig %v %v", c.Proxy, c.ProxyConfig.Type)`

sidecar/nginx/conf/amalgam8-services.conf.tmpl
```diff
@@ -0,0 +1,33 @@
+include /etc/nginx/amalgam8-access-logging.conf;
+include /etc/nginx/amalgam8-dynupstreams.conf;
+include /etc/nginx/nginx-customizations.conf;
+
+### A8 Sidecar proxy configuration
+server {
+       listen 6379{{.SSLFlag}};
+
+       ## Comment the previous line and enable the latter two if you are using the sidecar as a gateway
```
ijsnellf, line 9: We should probably remove these comments, since they imply you can manually modify this configuration file.

sidecar/nginx/conf/amalgam8-services.conf.tmpl
```diff
@@ -0,0 +1,33 @@
+include /etc/nginx/amalgam8-access-logging.conf;
```
ijsnellf, line 1: Add some sort of "Automatically generated file --don't modify" warning here?

sidecar/config/config.go
```diff
@@ -146,6 +159,9 @@ type Config struct {
 	Supervise bool `yaml:"supervise"`
 
 	HealthChecks []HealthCheck `yaml:"healthchecks"`
+	HealthChecksCACertPath string `yaml:"healthchecks_ca_cert_path"`
+
+	ProxyConfig ProxyConfig  `yaml:"proxyconfig"`
```
ijsnellf: "proxy_config"

sidecar/config/config.go
ijsnellf: I'm alright either way. Removing `A8_HEALTHCHECKS_CA_CERT_PATH` does make things simpler, but I can see why @paulben added it for convenience.

ijsnellf (changes requested):
Looks pretty good overall --a few comments about structure and style.

rshriram:
duplicate of #437

Issue #422, opened 2016-11-30
Zombie reaping issues


zcahana:
There are several fixes needed in the updated zombie reaping logic (copied from PR #421):

- The correct condition for enabling zombie reaping should be whether the sidecar runs as PID 1 or not, rather than inspecting a random environment variable.
- The parent process must forward any signal down to the child process.
- The parent process should exit with the same exit code of the child process.
- There's a fmt.Println() that slipped in and should probably go away (line 71).
- For the sake of code organization, let's not put all of this logic in the main file... we should strive to have that file only include configuration/construction code.

zcahana:
@GregHanson I assigned this to you given that you've been working on this. If you don't have time for this, please unassign.

zcahana:
Resolved by #423

PR #423, opened 2016-11-30
Zombie Cleanup Re-Org

GregHanson:
pull request resolving issues documented in issue #422 

 - moved the PID=1 logic out of sidecar.go
 - PID=1 propagates signals to child sidecar process
 - PID=1 process shuts down with the same exit code as the child sidecar process
 - uses syscall to see if PID=1 rather than presence of env var


sidecar/supervisor/init.go
```diff
@@ -0,0 +1,95 @@
...
+				continue
+			}
+			logrus.Debug("Zombie reaped")
+			// Done
+			break
+		}
+	}
+}
+
+func shutdown(sig int) {
```
ijsnellf: `func shutdown(code int)`?

ijsnellf (approved):

sidecar/supervisor/init.go
```diff
@@ -0,0 +1,95 @@
...
+	"os/signal"
+	"syscall"
+
+	"github.com/Sirupsen/logrus"
+)
+
+// Init launches new sidecar process with same args and env vars as current process
+// and runs zombie process cleanup for any inherited orphans
+func Init() {
+	cmd := exec.Command("a8sidecar", os.Args[1:]...)
```
zcahana: It would be more robust to use `exec.Command(os.Args[0], os.Args[1:]...)`

sidecar/supervisor/init.go
```diff
@@ -0,0 +1,95 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Env = os.Environ()
```
zcahana: No need to explicitly copy the environment variables, as by default they are inherited from the parent.

sidecar/sidecar.go
```diff
@@ -68,15 +66,9 @@ func Main() {
...
-		fmt.Println(os.Args[1:])
-		cmd := exec.Command("a8sidecar", os.Args[1:]...)
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-		cmd.Env = append([]string{"A8_CHILD=yes"}, os.Environ()...)
-		// listen for SIGCHLDs from any of the supervised processes and any orphans spun off from those
-		go supervisor.ReapZombies()
-		return cmd.Run()
+	// when PID=1, launch new sidecar process and assume init responsibilities
+	if os.Getpid() == 1 {
```
zcahana, line 70: Should this be an if-else condition, so that we have a distinct branches for being PID 1 or not?

sidecar/supervisor/init.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// Intercept SIGTERM/SIGINT and stop
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGTERM, syscall.SIGINT)
+	select {
+	case sig := <-sigChan:
+		logrus.Infof("Intercepted signal '%v'", sig)
+
+		// forwarding signal to supervised applications to exit gracefully
+		cmd.Process.Signal(sig)
```
zcahana: Race condition here: `cmd.Process` might be nil.

sidecar/supervisor/init.go
```diff
@@ -0,0 +1,95 @@
...
+	cmd.Env = os.Environ()
+
+	// listen for SIGCHLDs from any of the supervised processes and any orphans spun off from those
+	go reapZombies()
+
+	go launchChild(cmd)
+
+	// Intercept SIGTERM/SIGINT and stop
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGTERM, syscall.SIGINT)
```
zcahana: Forwarding SIGTERM/SIGINT once is not enough - we need to forward any signal (except for SIGCHLD), any time it's received, as long as the child process is alive, and exit only when the child exists.

I would structure the `Init()` method something along this line:
```go
func Init() {
	cmd := ... // build and set up command object
	
	err := cmd.Start()
	if err != nil {
		shutdown(exitCodeFromError(err))
	}
	
	// No we can launch the signal handler
	go handleSignals(cmd.Process)
	
	err = cmd.Wait()
	shutdown(exitCodeFromError(err))
}

func handleSignals(proc os.Process) {
	// Trap all signals
	sigChan := make(chan os.Signal, 1)
	signal.Notify(sigChan)
	
	for sig := range sigChan {
		switch sig {
			case syscall.SIGCHLD:
				reapZombie()
			default:
				forwardSignal(proc, sig)
		}
	}
}
```

zcahana (changes requested):

sidecar/sidecar.go
GregHanson: coding convention that golint checks against, it complains when there are return statements in an if/else
```
if PID==1 {

    //do something
    return value
} else {
    //do something else
    return value
}
```
specifically: `amalgam8/sidecar/sidecar.go:75:9: if block ends with a return statement, so drop this else and outdent its block`

ijsnellf (approved):
LGTM

zcahana:
I've just realized that we redo issue #401 here, but this time there's a simple solution: instead of calling `cmd.Wait()` (and racing with the SIGCHLD handler), we can inspect the return value of `syscall.Wait4()` to get the PID of the waited-on child, and compare it to `proc.Pid`. If they match, we can `shutdown()`.

GregHanson:
@zcahana you are right, we would hit that problem only on exit.  I guess the expected race condition is that if SIGCHLD causes us to read the return code before the `cmd.Wait()` does, the PID=1 process would end up exiting with a bad return code.  Based on your comment, is this what you were suggesting?
```func reapZombies(proc *os.Process) {
	for {
		pid, err := syscall.Wait4(-1, nil, syscall.WNOHANG, nil)

		if proc != nil && proc.Pid == pid {
			shutdown(err)
		}
		
		// Spurious wakeup
		if err == syscall.EINTR {
			continue
		}
		logrus.Debug("Zombie reaped")
		// Done
		break
	}
}```

zcahana (approved):

zcahana:
LGTM-too. Merging.

PR #425, opened 2016-12-05
Registry API and adapters packages reorg

zcahana:
Move the registry-related interfaces and structs out of the registry package, so that the non-a8-registry implementations (k8s, eureka) wouldn't be coupled to the a8-registry.
This is the same approach we'll be taking with the rules-related interfaces and structs, so that they won't be coupled into the a8 controller.


zachidan (approved):

rshriram:
Can you please do the same for controller as well, as part of this PR? I doubt it would be a big change. I am sure Greg and Isaiah would be happy with it.

zcahana:
@rshriram this is already being worked on by @zachidan. Will be submitted through a different PR though.

PR #429, opened 2016-12-06
Added gremlin to example yamls for k8s and docker-compose

### LINKED ISSUE #381: Need ability to call gremlin as a service from the new GO based CLI
Another item associated with supporting the new Go CLI is the requirement to call Gremlin as a service. There's a python server here: https://github.com/amalgam8/gremlin/blob/master/app.py	
GitHub
amalgam8/gremlin

Somebody needs to create a docker image, built it, and make the changes to the controlplane startup yaml files to run it.  

In order for the 'a8ctl recipe-run' command to runn successfully, it will need to have the Gremlin server up and running in the controlplane before it can be tested.

JacobFrericks:
Fixes #381


rshriram:
Do not merge yet. Depends on a8ctl Go cli going into master.

PR #430, opened 2016-12-07
Controller API package reorg

zachidan:
Move the controller-related interfaces and structs out of the controller package, so that the non-a8-controller implementations (k8s) wouldn't be coupled to the a8-controller.


controller/api/rules.go
```diff
@@ -23,13 +23,14 @@ import (
 	"github.com/amalgam8/amalgam8/controller/metrics"
 	"github.com/amalgam8/amalgam8/controller/rules"
 	"github.com/amalgam8/amalgam8/controller/util/i18n"
+	"github.com/amalgam8/amalgam8/pkg/api"
 	"github.com/ant0ine/go-json-rest/rest"
 )
 
 // RuleList is used to output the results of rule queries.
 type RuleList struct {
```
zcahana: Is `RuleList` still needed? Can we use `api.RuleSet` instead?

pkg/api/filter.go
```diff
@@ -0,0 +1,30 @@
...
+
+	// RuleRoute denotes rules with a routing field
+	RuleRoute
+
+	// RuleAction denotes rules with an action field
+	RuleAction
+)
+
+// Filter to apply to sets of rules.
+type Filter struct {
```
zcahana: A more descriptive name would be `api.RuleFilter`.
`rules.Filter` had the context in the package name, which is lost here.

controller/rules/validation.go
```diff
@@ -19,12 +19,13 @@ import (
 	"fmt"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/amalgam8/amalgam8/pkg/api"
 	"github.com/xeipuuv/gojsonschema"
 )
 
 // Validator validates rules against the rule schema.
 type Validator interface {
```
zcahana, line 26: It seems that the rule validation logic should go out of the controller packages too, because it is supposed to be independent of any RuleStore implementation such as the A8 controller.

controller/rules/filter.go
```diff
@@ -15,51 +15,11 @@
...
-}
-
-// String representation of the filter
-func (f Filter) String() string {
-	return fmt.Sprintf("%#v", f)
-}
-
 // FilterRules returns a list of filtered rules.
-func FilterRules(f Filter, rules []Rule) []Rule {
+func FilterRules(f api.Filter, rules []api.Rule) []api.Rule {
```
zcahana: I think this function should go into `pkg/api/filter.go` as well, because it defines the interaction between an `api.Filter` and an `api.Rule`, which is supposed to be independent of the a8-controller implementation.

And that would be a good change to make it a method of `api.Filter`:
```go
func (f *Filter) Apply(rules []api.rule) []api.Rule { ... }
```

controller/rules/filter_test.go
```diff
@@ -17,6 +17,8 @@ package rules
 import (
 	"reflect"
 	"testing"
+
```
zcahana: Ditto, should go into `/pkg/api`.

sidecar/config/config.go
```diff
@@ -139,9 +125,14 @@ type Config struct {
 	Service  Service  `yaml:"service"`
 	Endpoint Endpoint `yaml:"endpoint"`
 
-	Registry   Registry   `yaml:"registry"`
-	Controller Controller `yaml:"controller"`
-	Dnsconfig  Dnsconfig  `yaml:"dnsconfig"`
+	DiscoveryType string `yaml:"discovery_type"`
+	RulesType     string `yaml:"rules_type"`
```
zcahana: Despite our discussion yesterday, I think that type is not the best term to use here.
`rules_type: amalgam8` or `rules_type: kubernetes` sounds a bit weird. `discovery_type` somehow makes more sense.
All in all, I think `rules_backend: kubernetes` and `discovery_backend: kubernetes` would be a better fit.

sidecar/config/config.go
```diff
@@ -66,48 +66,34 @@ type Endpoint struct {
 	Type string `yaml:"type"`
 }
 
-// Registry configuration
-type Registry struct {
```
zcahana, line 70: Despite our discussion yesterday, I think we should keep (and deprecate) the top-level `registry:` and `controller:` for backwards compatibility. I'm aware of at least two users who are using config files with the sidecar, which would be broken by this.

sidecar/config/default.go
```diff
@@ -32,33 +32,32 @@ var DefaultConfig = Config{
...
-	Controller: Controller{
-		URL:   "",
-		Token: "",
-		Poll:  time.Duration(15 * time.Second),
+	RulesType: Amalgam8Type,
+
+	Kubernetes: Kubernetes{
+		URL:       "",
+		Token:     "",
+		Namespace: "",
```
zcahana: Should we use "default" namespace as the default?

sidecar/config/flags.go
```diff
@@ -136,21 +137,25 @@ var Flags = []cli.Flag{
...
 	cli.StringFlag{
+		Name:   rulesTypeFlag,
+		EnvVar: envVar(rulesTypeFlag),
+		Usage:  fmt.Sprintf("Rules service type (%s, %s)", Amalgam8Type, KubernetesType),
+	},
+	cli.StringFlag{
 		Name:   controllerURLFlag,
 		EnvVar: envVar(controllerURLFlag),
-		Usage:  "URL for Controller service",
+		Usage:  "URL for Rules service",
```
zcahana: The `--controller_XXX` usage strings should explicitly indicate that this is the Amalgam8 Controller, not "Rules service" or "Amalgam8 Rules".

sidecar/config/config_test.go
```diff
@@ -58,6 +58,7 @@ var _ = Describe("Config", func() {
 			Expect(c.Endpoint.Type).To(Equal(DefaultConfig.Endpoint.Type))
 			Expect(c.DiscoveryType).To(Equal(DefaultConfig.DiscoveryType))
 			Expect(c.RulesType).To(Equal(DefaultConfig.RulesType))
+			Expect(c.RulesType).To(Equal(DefaultConfig.RulesType))
```
zcahana: Duplicate assertion.

zcahana (changes requested):

controller/rules/validation.go
rshriram: I agree

zcahana (approved):

zcahana:
LGTM. @zachidan Please fix merge conflict and we can merge.

ijsnellf (commented):
LGTM.

steveclay:
@rshriram I'm a little confused by this item as it seems to say that the a8 controller is not always in the picture (e.g. when k8s is used).  That is different than the understanding my team has had from past conversations where it was said that the a8 registry is swapped out with a registry of k8s, but that the a8 controller is always in use.  Can you provide some explanation so we have the right understanding?  Thanks

rshriram:
In k8s, the controller's functionality would be pushed down to the sidecar (rule validation, config generation, etc.). We are trying to decouple a whole bunch of things.. how the rules are received by the container runtime/a8 (whether through a8 APIs or through kubernetes native apis), how and when and where the rules are processed/validated (at a special controller service or at sidecar level) and how/when/where the config for sidecars are generated.

From a usage point of view, the idea is that in k8s, the user would use native kubectl tool to provide rules as Thirdparty resources (kubernetes.io/docs/user-guide/thirdpartyresources/). Some entity in the backend (abstracted from user) will pick up the rules and validate them, and then apply them to all sidecars.

So, all in all, the data store for rules is k8s itself, instead of relying on Redis. The only moving piece in this picture would be the sidecar.

PR #431, opened 2016-12-07
A8controller cache

shlomiti:
This PR adds a cache for rules fetching. Sidecar main now uses this cache for A8. This is similar to the caching mechanism we have for the A8 registry fetch.

### DIFF
--- a/controller/client/cache.go
+++ b/controller/client/cache.go
@@ -1,0 +1,94 @@
+// Copyright 2016 IBM Corporation
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


rshriram:
This is unrelated, and the previous PR got merged before I could respond. Please remove the *-wip.json from the api package.. We don't need it.

rshriram (changes requested):
Can we move the cache to the pkg/ as its generic for different rules backends.

zcahana:
@rshriram the `*-wip.json` is indeed unrelated. I removed it in a separate PR #435.

zcahana:
Also, regarding the generic cache - this also applies to the a8 registry caching client, which can be generic in the same manner. Given that we don't want to mix registry stuff in this PR, I'll submit another one with the proposed changes.

zcahana:
Otherwise, LGTM. I think this can be merged as-is.

rshriram (approved):

zachidan:
The cache.ListRules method should skip the api.FilterRules call if the provided filter is either nil or empty.

zcahana:
@zachidan addressed in #440.

PR #434, opened 2016-12-07
Proxy adapter interface

ijsnellf:
This PR moves the sidecar's proxy management behind an abstract interface so that different implementations (NGINX, Envoy) can be specified dynamically.

This is a work in progress, but comments are welcome.


sidecar/proxy/adapter.go
```diff
@@ -0,0 +1,25 @@
...
+package proxy
+
+import "github.com/amalgam8/amalgam8/pkg/api"
+
+// Adapter interface for proxy implementations.
+type Adapter interface {
+	Start() error
+	Stop() error
+
+	GetState() ([]api.ServiceInstance, []api.Rule)
```
rshriram: Could you please add some comments for these functions?

sidecar/config/config.go
```diff
@@ -119,9 +125,10 @@ type HealthCheck struct {
 
 // Config stores the various configuration options for the sidecar
 type Config struct {
-	Register bool `yaml:"register"`
-	Proxy    bool `yaml:"proxy"`
-	DNS      bool `yaml:"dns"`
+	Register     bool   `yaml:"register"`
+	Proxy        bool   `yaml:"proxy"`
+	ProxyAdapter string `yaml:"proxy_adapter"`
```
rshriram, line 142: Can we move the proxyadapter under Proxy? as say, a "type" field ? This would allow us to add additional fields such as the port on which the proxy should be configured to listen on (something that we do not auto-configure now), and some other common proxy related settings.

sidecar/config/config.go
```diff
@@ -358,7 +366,10 @@ func (c *Config) Validate() error {
 	}
 
 	if c.Proxy {
-		validators = append(validators, IsInSet("Rules service backend", c.RulesBackend, []string{Amalgam8Backend, KubernetesBackend}))
+		validators = append(
+			validators,
+			IsInSet("Rules service backend", c.RulesBackend, []string{Amalgam8Backend, KubernetesBackend}),
```
rshriram, line 389: Rename Amalgam8Backend to StandaloneBackend (signifying that the controller is just a standalone instance in this case).

sidecar/proxy/adapter.go
ijsnellf: Will do.

sidecar/config/config.go
ijsnellf: Sure. That naming scheme was added by https://github.com/amalgam8/amalgam8/pull/430.

sidecar/proxy/monitor/discovery.go
```diff
@@ -64,6 +71,21 @@ func NewDiscoveryMonitor(conf DiscoveryConfig) Monitor {
 	}
 }
 
+// Not safe if monitor has started
+func (m *discoveryMonitor) AddListener(listener DiscoveryListener) {
+	m.listeners = append(m.listeners, listener)
+}
```
GregHanson, line 87: Should probably place a lock around the append state

sidecar/proxy/monitor/discovery.go
```diff
@@ -64,6 +71,21 @@ func NewDiscoveryMonitor(conf DiscoveryConfig) Monitor {
...
+
+// Not safe if monitor has started
+func (m *discoveryMonitor) RemoveListener(listener DiscoveryListener) {
+	for i := range m.listeners {
+		if m.listeners[i] == listener {
+			m.listeners = append(m.listeners[:i], m.listeners[i+1:]...)
```
GregHanson: same comment as above

sidecar/proxy/monitor/rules.go
```diff
@@ -53,27 +60,42 @@ func NewRulesMonitor(conf RulesConfig) Monitor {
 	}
 }
 
+// Not safe if monitor has started
+func (m *rulesMonitor) AddListener(listener RulesListener) {
+	m.listeners = append(m.listeners, listener)
```
GregHanson: same lock comment as above

sidecar/proxy/monitor/rules.go
```diff
@@ -53,27 +60,42 @@ func NewRulesMonitor(conf RulesConfig) Monitor {
...
+}
+
+// Not safe if monitor has started
+func (m *rulesMonitor) RemoveListener(listener RulesListener) {
+	for i := range m.listeners {
+		if m.listeners[i] == listener {
+			m.listeners = append(m.listeners[:i], m.listeners[i+1:]...)
```
GregHanson: lock

GregHanson (changes requested):

sidecar/config/config.go
ijsnellf: I think that is a better way to organize the config in general, but it is also a breaking change. The entire config could be reworked, so I think it might be a better item for a different issue/PR.

GregHanson (approved):

ijsnellf:
This should be ready to be merged once it is reviewed.

sidecar/config/config_test.go
```diff
@@ -406,9 +406,10 @@ log_level: debug
 					Port:   8053,
 					Domain: "amalgam8",
 				},
-				Proxy:    true,
-				Register: true,
-				DNS:      true,
+				Proxy:        true,
+				ProxyAdapter: NGINXAdapter,
```
zcahana, line 457: Should add the ProxyAdapter to all config test cases (env vars, flags YAML, defaults ...)

sidecar/config/flags.go
```diff
@@ -69,7 +70,12 @@ var Flags = []cli.Flag{
 	cli.BoolFlag{
 		Name:   proxyFlag,
 		EnvVar: envVar(proxyFlag),
-		Usage:  "Enable automatic service discovery and load balancing across services using NGINX",
+		Usage:  "Start proxy",
+	},
+	cli.StringFlag{
+		Name:   proxyAdapterFlag,
+		EnvVar: envVar(proxyAdapterFlag),
+		Usage:  "Proxy adapter (nginx, envoy)",
```
zcahana: How about `fmt.Sprintf("Proxy adapter (%s, %s)", NGINXAdapter, EnvoyAdapter)`?

sidecar/config/flags.go
```diff
@@ -25,6 +25,7 @@ const (
 	configFlag              = "config"
 	registerFlag            = "register"
 	proxyFlag               = "proxy"
+	proxyAdapterFlag        = "proxy_adapter"
```
zcahana, line 32: I'm wondering if we need to rename `--discovery_backend` and `--rules_backend` to `--discovery_adapter` and `--rules_adapter`. Consistency is never a bad thing, and `--proxy_backend` doesn't sound right. Thoughts?

sidecar/proxy/monitor/discovery.go
```diff
@@ -64,6 +75,25 @@ func NewDiscoveryMonitor(conf DiscoveryConfig) Monitor {
 	}
 }
 
+// AddListener adds a listener
+func (m *discoveryMonitor) AddListener(listener DiscoveryListener) {
+	m.mutex.Lock()
+	m.listeners = append(m.listeners, listener)
```
zcahana: Any reason why not use the lock-defer-unlock idiom?

sidecar/proxy/monitor/discovery.go
```diff
@@ -64,6 +75,25 @@ func NewDiscoveryMonitor(conf DiscoveryConfig) Monitor {
...
+	m.mutex.Unlock()
+}
+
+// RemoveListener removes the listener
+func (m *discoveryMonitor) RemoveListener(listener DiscoveryListener) {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	for i := range m.listeners {
```
zcahana, line 101: This code is a bit fragile, because the concrete `DiscoveryListener`'s compared might not be [comparable](https://golang.org/ref/spec#Comparison_operators), which will result with a runtime panic. I'd suggest that by minimum, we enclose the comparison with panic recovery code.

sidecar/proxy/monitor/discovery.go
```diff
@@ -113,6 +143,8 @@ func (m *discoveryMonitor) poll() error {
 
 	// Notify the listeners
 	instancesByValue := instanceListAsValues(instances)
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
```
zcahana: Invoking callbacks with a mutex acquired is generally not a good idea (more prone to deadlocks, long running operations, ..). 

Actually, a previous incarnation of the `RegistryMonitor` had to deal with this problem as well, and used copy-on-write method for that: When a listener is added/removed, a _new_ slice is allocated, populated, and assigned to `m.listeners`. When a callback is invoked, the slice _pointer_ is atomically copied from m.listeners, and then the callbacks are invoked while iterating on the local slice pointer, without a lock.

sidecar/proxy/nginxadapter.go
```diff
@@ -0,0 +1,97 @@
...
+	return a.service.Stop()
+}
+
+// CatalogChange updates on a change in the catalog.
+func (a *NGINXAdapter) CatalogChange(instances []api.ServiceInstance) error {
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+
+	a.instances = instances
+	return a.manager.Update(a.instances, a.rules)
```
zcahana, line 92: Need to consider if we want to invoke `a.manager.Update(a.instances, a.rules)` with the mutex held, or just queue that operation (e.g., in a chan) for a processing by another goroutine.

sidecar/sidecar.go
```diff
@@ -260,56 +260,45 @@ func buildServiceRules(conf *config.Config, kubeClient kubernetes.Interface) (ap
...
+		logrus.WithError(err).Error("Could not start proxy adapter")
+		return err
+	}
 
-	debugger := debug.NewAPI(nginxProxy)
+	debugger := debug.NewAPI()
+	rulesMonitor.AddListener(debugger)
+	discoveryMonitor.AddListener(debugger)
 
 	a := rest.NewApi()
```
zcahana, line 303: Unrelated to this PR, but the configuration of the debug API server (lines 303-319) should probably go into the `debug` package (e.g., as part of `debug.NewAPI()`).

zcahana (changes requested):

sidecar/config/flags.go
rshriram: +1

ijsnellf: Makes sense.

sidecar/proxy/monitor/discovery.go
ijsnellf: Nope, no reason. I can switch to the defer idiom.

sidecar/proxy/monitor/discovery.go
ijsnellf: Good point. In our case, the implementations of DiscoveryListeners are always pointers, which are comparable. Do you think it would be sufficient to put a warning on DiscoveryListener that the implementation is comparable? This seems like a programmer error.

sidecar/sidecar.go
ijsnellf: Definitely. Wanted to keep this PR focused, though.

sidecar/config/flags.go
GregHanson: Since renaming a flag/env var is a breaking change, I opened an issue for this so we can submit an additional PR for this :
https://github.com/amalgam8/amalgam8/issues/446

sidecar/config/flags.go
ijsnellf: Done.

sidecar/proxy/monitor/discovery.go
zcahana: Yes, this is probably sufficient in our case.

sidecar/proxy/monitor/discovery.go
```diff
@@ -142,10 +157,16 @@ func (m *discoveryMonitor) poll() error {
 	m.cache = catalog
 
 	// Notify the listeners
+	var listeners []DiscoveryListener
+	func() {
+		m.mutex.Lock()
+		defer m.mutex.Unlock()
+		listeners = make([]DiscoveryListener, len(m.listeners))
```
zcahana, line 164: It is not needed to allocate `listeners` and copy `m.listeners` into it. It is sufficient (and correct) to just copy the `m.listeners` _reference_ into `listeners`. Using copy-on-write during add/remove guarantees that the slice referenced by `listeners` is immutable, so can be iterated-on without a lock held. However, it's still important to copy the slice reference with a lock held, both for atomicity and [happens-before](https://golang.org/ref/mem#tmp_2) relation.

sidecar/proxy/monitor/rules.go
```diff
@@ -129,9 +143,15 @@ func (m *rulesMonitor) poll() error {
...
 
-	// Notify listeners
-	m.mutex.Lock()
-	defer m.mutex.Unlock()
+	// Notify the listeners
+	var listeners []RulesListener
+	func() {
+		m.mutex.Lock()
+		defer m.mutex.Unlock()
+		listeners = make([]RulesListener, len(m.listeners))
```
zcahana, line 151: Ditto.

sidecar/config/flags.go
zcahana: Renaming `--discovery_backend` and `--rules_backend` is _not_ a breaking change because these flags were never released. Both are post-[v0.4.2](https://github.com/amalgam8/amalgam8/blob/v0.4.2/sidecar/config/flags.go).
Either way, let's make sure to resolve #446 before the next release, and with a different PR.

sidecar/proxy/adapter.go
```diff
@@ -14,14 +14,8 @@
...
-
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-)
-
-func TestPackage(t *testing.T) {
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "Poll Suite")
+// Adapter interface for proxy implementations.
+type Adapter interface {
```
zcahana, line 18: Have you considered extending the `Adapter` so that it embeds both `DiscoveryListener` and `RulesListener`?
```golang
type Adapter interface {
	monitor.DiscoveryListener
	monitor.RulesListener

 	Start() error
 	Stop() error
}
```

We anyway expect all proxy adapters to be somehow able to react to discovery/rules changes. This also decouples the various proxy adapters from `DiscoveryMonitor` and `RulesMonitor`. And as a bonus, we infuse some more meaning to the proxy adapter abstraction, other than "something that can be Start()'ed and Stop()'ed".

These are not major benefits, but maybe somewhat cleaner design.

sidecar/proxy/monitor/discovery.go
ijsnellf: Makes sense.

sidecar/proxy/adapter.go
ijsnellf: Yes, I was on the fence about that design choice. However, it will work well for both the Envoy and NGINX adapters, so that seems like a good simplification.

zcahana (approved):

rshriram:
@ijsnellf @GregHanson ^^^.

PR #436, opened 2016-12-08
Improved NGINX process management

ijsnellf:
This PR attempts to make NGINX process management more robust. Our existing NGINX process management uses OS specific commands and does not properly ensure that the NGINX process is running. To remedy this I redesigned the code to use standard Go libraries and automatically restart the service in the case of unexpected exits.


rshriram (commented):
Can we abstract this across proxies and not just nginx?

ijsnellf:
@rshriram: Maybe as part of #434 or later Envoy work?

GregHanson (approved):
LGTM

PR #437, opened 2016-12-08
Sidecar TLS support

paulben:
Secure the following paths:
- service->nginx->service
- sidecar->service (for healthcheck)

New proxy_config.go is called from sidecar to configure proxy. It is intended to be generic enough to easily handle envoy.


sidecar/config/config.go
```diff
@@ -140,6 +149,8 @@ type Config struct {
 
 	HealthChecks []HealthCheck `yaml:"healthchecks"`
 
+	ProxyConfig ProxyConfig `yaml:"proxyconfig"`
```
ijsnellf, line 152: `yaml:"proxy_config"` to keep the naming scheme consistent.

sidecar/config/proxy_config.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"os"
+	"strconv"
+	"text/template"
+
+	"github.com/Sirupsen/logrus"
+)
+
+type a8ServicesNginxConfData struct {
```
ijsnellf: Prefixing structs and functions with `a8` or `Amalgam8` inside the Amalgam8 repository is unnecessary and can make code harder to read. Could you refactor accordingly?

sidecar/config/proxy_config.go
```diff
@@ -0,0 +1,88 @@
...
+// GenProxyConfig generates proxy config
+func (c *Config) GenProxyConfig() error {
+	// Generate amalgam8-services.conf
+	err := c.genNginxA8ServicesConf()
+	return err
+}
+
+func (c *Config) genNginxA8ServicesConf() error {
+	var a8ServicesData *a8ServicesNginxConfData
+	logrus.Debug("genNginxA8ServicesConf " + strconv.FormatBool(c.Proxy) + " " + strconv.FormatBool(c.ProxyConfig.TLS))
```
ijsnellf: Remove or replace with a more informative message of the form `logrus.Debug("Generating services configuration.")`

sidecar/config/proxy_config.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	file, err := os.Create(targetFile)
+	if err != nil {
+		return err
+	}
+	err = goTemplate.Execute(file, data)
+	if err != nil {
```
ijsnellf: Style tip:
`if err := goTemplate.Execute(file, data); err != nil {`

sidecar/config/proxy_config.go
```diff
@@ -0,0 +1,88 @@
...
+		return err
+	}
+	if _, err := os.Stat(targetFile); !os.IsNotExist(err) {
+		err := os.Remove(targetFile)
+		if err != nil {
+			return err
+		}
+	}
+	file, err := os.Create(targetFile)
```
ijsnellf, line 79: Lines 73-78 are unnecessary because `os.Create(targetFile)` will overwrite any existing files. See: https://golang.org/pkg/os/#Create

sidecar/nginx/conf/amalgam8-services.conf.tmpl
```diff
@@ -0,0 +1,33 @@
+include /etc/nginx/amalgam8-access-logging.conf;
```
ijsnellf, line 1: @rshriram: Should the comments about how to configure this file be removed, since this entire config file is automatically generated?

sidecar/sidecar.go
```diff
@@ -233,14 +239,10 @@ func buildServiceDiscovery(conf *config.Config) (api.ServiceDiscovery, error) {
...
 	switch strings.ToLower(conf.RulesBackend) {
 	case config.Amalgam8Backend:
-		controllerConf := controllerclient.CacheConfig{
-			Config: controllerclient.Config{
-				URL:       conf.A8Controller.URL,
-				AuthToken: conf.A8Controller.Token,
-			},
-			PollInterval: conf.A8Controller.Poll,
-		}
-		return controllerclient.NewCache(controllerConf)
```
ijsnellf: This looks like an accidental change. The change reverts this code to an earlier version. Please remove this change.

sidecar/register/healthcheck/http.go
```diff
@@ -45,12 +48,32 @@ func NewHTTP(conf config.HealthCheck) (Check, error) {
 	if err := validateHTTPConfig(&conf); err != nil {
 		return nil, err
 	}
-
+	var client *http.Client
+	//if conf.UpstreamTrustPath != "" {
```
ijsnellf, line 52: Remove this commented out code.

sidecar/config/proxy_config.go
```diff
@@ -0,0 +1,88 @@
...
+	A8ServicesConfTemplatePath        = "/etc/nginx/amalgam8-services.conf.tmpl"
+	A8ServicesConfPath                = "/etc/nginx/amalgam8-services.conf"
+	NginxSSLFlag                      = " ssl"
+	NginxSSLCertDirective             = "ssl_certificate "
+	NginxSSLCertKeyDirective          = "ssl_certificate_key "
+	NginxProxySSLTrustedCertDirective = "proxy_ssl_trusted_certificate "
+)
+
+// GenProxyConfig generates proxy config
+func (c *Config) GenProxyConfig() error {
```
ijsnellf, line 43: The generation code in this file should be moved to `sidecar/proxy/nginx` and either be standalone functions or part of a struct or interface, not member functions of `sidecar/config/Config`.

sidecar/config/flags.go
```diff
@@ -77,6 +81,26 @@ var Flags = []cli.Flag{
 		Usage:  "Enable DNS server",
 	},
 	cli.StringFlag{
```
ijsnellf: Probably a better candidate for `cli.BoolFlag`.

sidecar/sidecar.go
```diff
@@ -77,6 +77,12 @@ func sidecarCommand(context *cli.Context) error {
 	if err != nil {
 		return err
 	}
+	if conf.Proxy {
+		err = conf.GenProxyConfig()
+		if err != nil {
+			return err
+		}
+	}
```
ijsnellf: The logic for generating the initial NGINX config should be somewhere in the `Run` function, not at the top level here in `sidecarCommand`.

ijsnellf (changes requested):
Some comments and requests...

sidecar/config/config.go
paulben: Consistent with what? I modeled it after dnsconfig.

sidecar/config/proxy_config.go
paulben: I agree that in general an a8 prefix is unnecessary. In this case the a8Services prefix is specifically intended to be descriptive of the target file: amalgam8-services.conf.

sidecar/config/proxy_config.go
paulben: Yep, good point.

sidecar/sidecar.go
paulben: Oops. Good catch. Thank you.

sidecar/register/healthcheck/http.go
paulben: Yep.

sidecar/config/flags.go
paulben: Indeed.

sidecar/sidecar.go
paulben: I'm curious where the envoy config is being generated? If that's in `Run` then the nginx config should be generated there also. 

If the place for envoy config generation is not yet detmined then: given that this is the 1st dynamic generation of nginx config it seems to me to be a subjective/philosphical call: 
- Is proxy config generation part of the part of pre-Run config processing? 
- Or is it part of the Run phase?

Functionally the outcome seems identical.

sidecar/config/proxy_config.go
paulben: Can't say I disagree. As I'm not involved in the envoy design/code it's hard for me to know the best way to fit this in. 

Could we start with this and then you guys decide the best place/structure for it in the context of the envoy changes? And you refactor as appropriate? 

It seems useful to get this in the stream sooner rather than later to avoid more merging delays/commits, particularly given that this is a question of optimal code structure rather than a functional issue.

sidecar/config/proxy_config.go
paulben: Ok.

sidecar/nginx/conf/amalgam8-services.conf.tmpl
```diff
@@ -0,0 +1,33 @@
...
+
+### A8 Sidecar proxy configuration
+server {
+       listen 6379{{.SSLFlag}};
+
+       ## Comment the previous line and enable the latter two if you are using the sidecar as a gateway
+       #listen 80;
+       #listen 443 ssl;
+
+       ## For using the A8 proxy as a gateway with SSL support
```
rshriram: Remove these comments please

sidecar/nginx/conf/amalgam8-services.conf.tmpl
```diff
@@ -0,0 +1,33 @@
...
+       # eventhough the key and cert are stored in same file, only cert is
+       # sent to client
+       {{.CertPath}}
+       {{.CertKeyPath}}
+
+       location ~ ^/(?<service_name>.*?)(?<reqpath>/(.*))?$ {
+         include /etc/nginx/amalgam8-rules.conf;
+         # Add proxy SSL verification directives here:
```
rshriram, line 21: Change this comment to 
```nginx
####DO NOT CHANGE#####
{{.CACertPath}}
######################
```

Apply the same style to other parts of the config that have the template code.

rshriram (changes requested):

sidecar/config/proxy_config.go
paulben: Would that be the "...truncating it if it already exists." part in the referenced doc?

I'll code it up and give it a shot given that the file will exist as the product is today.

sidecar/config/proxy_config.go
rshriram: Lets please avoid using a8 inside the code. The code is not bound to that filename. You can rename that file and fix nginx.conf and things would stil work.

sidecar/nginx/conf/amalgam8-services.conf.tmpl
paulben: Done.

sidecar/sidecar.go
paulben: Done.

sidecar/sidecar.go
paulben: Moved to Run.

sidecar/nginx/conf/amalgam8-services.conf.tmpl
paulben: Done (see @rshriram's comment below)

sidecar/config/proxy_config.go
paulben: Done.

sidecar/config/proxy_config.go
paulben: Done.

sidecar/config/proxy_config.go
paulben: Done.

sidecar/config/proxy_config.go
paulben: Done.

sidecar/register/healthcheck/http.go
paulben: Done.

sidecar/config/flags.go
paulben: Done.

sidecar/sidecar.go
rshriram: Where ? I still see it in the diff.

sidecar/config/proxy_config.go
rshriram: I don't see it in the diff. Perhaps you forgot to commit?

rshriram (dismissed):
Two comments (things you probably forgot to commit).

sidecar/config/proxy_config.go
ijsnellf: > Could we start with this and then you guys decide the best place/structure for it in the context of the envoy changes? And you refactor as appropriate?

My comment above suggests the appropriate place to put the functions/structure. If you want this PR to be merged, please change it.

ijsnellf (dismissed):

PR #441, opened 2016-12-11
Extract Kubernetes utilities into 'pkg/kubernetes' package

zcahana:
So that it can be re-used across the discovery adapter, rules adapter, and services/labels auto-inference mechanism.


pkg/kubernetes/client.go
```diff
@@ -0,0 +1,53 @@
...
+	URL   string
+	Token string
+}
+
+// NewClient creates a new Kubernetes client using the specified configuration.
+// If no URL and Token are specified, then these values are attempted to be read from the
+// service account (if running within a Kubernetes pod).
+func NewClient(config Config) (kubernetes.Interface, error) {
+	var kubeConfig *rest.Config
+	if config.URL != "" || config.Token != "" {
```
zachidan: This condition should be -

if config.URL != "" {
    // Use custom client

zachidan (changes requested):

pkg/kubernetes/client.go
zcahana: Done.

zachidan (approved):

PR #447, opened 2016-12-14
Kubernetes Rules contoller

zachidan:
This PR adds a kubernetes controller that validates new/updated rule resources.

Rules are ThirdPartyResources which are stored internally in kubernetes. Rules can be added either by using the **kubectl** command line or by interacting with the kubernetes API directly.


rshriram (changes requested):
is this a work in progress? Didn't we decide to not have a controller at all, instead let the user create the TPR and sidecars validate the content?

I am fine with this approach as well. Except this needs to integrate with the rest of the controller logic and send the rules to the sidecars.

Makefile
```diff
@@ -41,14 +41,17 @@ APP_VER		:= $(shell git describe 2> /dev/null || echo "unknown")
 REGISTRY_APP_NAME		:= a8registry
 CONTROLLER_APP_NAME		:= a8controller
 SIDECAR_APP_NAME		:= a8sidecar
+K8SRULES_APP_NAME		:= a8k8srulescontroller
```
rshriram, line 45: Do we need a separate binary? Why can't we merge this into the current A8 controller ?

cmd/k8srules/controller.go
```diff
@@ -0,0 +1,273 @@
+// Copyright 2016 IBM Corporation
+//
```
rshriram, line 2: Same comment as above. If we add a command line flag to the controller that it is running as a K8S service, then it could automatically switch to using this code instead of the generic code.

zachidan:
@rshriram The controller sets a special flag on the TPR with the result of the validation: status.state = valid|invalid. The sidecar will retrieve ONLY rules with status.state = valid.

The kubernetes rules adapter will be added in a separated PR.

cmd/k8srules/controller.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// rulesCache caches service rules resources from Kubernetes API
+	rulesCache      cache.Store
+	rulesController cache.ControllerInterface
+
+	// workqueue is used to queue and process events send from the cache controllers
+	workqueue *kubepkg.Workqueue
+
+	// validator validates rules against the rule schema
+	validator a8api.Validator
```
rshriram: No A8 references inside the code please

cmd/k8srules/controller.go
```diff
@@ -0,0 +1,273 @@
...
+
+	err := c.validator.Validate(rule.Spec)
+	if err != nil {
+		log.Printf("Rule %#v is not valid. %s\n", rule, err)
+		updRule.Status.State = kuberules.RuleStateInvalid
+		updRule.Status.Message = err.Error()
+	} else {
+		log.Printf("Rule %s is valid\n", rule.Metadata.Name)
+		updRule.Status.State = kuberules.RuleStateValid
+	}
```
rshriram, line 132: If a sidecar is watching this TPR, will it receive the entire object or just the delta?

rshriram: By delta, I mean will it be just that line in JSON or will the sidecar receive the entire Rules Array that is stuffed into the TPR? Any chance that this can be compressed ?

cmd/k8srules/controller.go
```diff
@@ -0,0 +1,273 @@
...
+	log.Printf("Rule object updated: %s (state: %s, version: %s -> %s)\n",
+		newRule.Metadata.Name,
+		newRule.Status.State,
+		oldRule.Metadata.ResourceVersion,
+		newRule.Metadata.ResourceVersion)
+
+	if newRule.Status.State == "" {
+		c.validateRule(newRule)
+	}
+}
```
rshriram, line 195: What about delete? When a rule gets deleted, I presume that the sidecar gets notified? Once again, is it the delta or is it the entire rules array minus the deleted item?

cmd/k8srules/rc.yaml
```diff
@@ -0,0 +1,25 @@
...
+    spec:
+      containers:
+        - name: rules
+          image: amalgam8/a8-k8s-rules-controller:latest
+          imagePullPolicy: IfNotPresent
+          env:
+            - name: K8S_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
```
rshriram, line 25: See first comment. We need to merge this into controller such that controller switches to this operating mode when provided with K8S env vars/flags

docker/Dockerfile.k8srules
```diff
@@ -0,0 +1,21 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+FROM alpine:3.4
+
+WORKDIR /opt/a8k8srulescontroller
+COPY bin/a8k8srulescontroller    /opt/a8k8srulescontroller/
+COPY pkg/api/rules-schema.json   /opt/a8k8srulescontroller/rules-schema.json
+
+ENTRYPOINT [ "/opt/a8k8srulescontroller/a8k8srulescontroller" ]
```
rshriram, line 21: Kill, as per previous comment

pkg/adapters/rules/kubernetes/rule.go
```diff
@@ -0,0 +1,117 @@
+// Copyright 2016 IBM Corporation
```
rshriram: Can we call this routingrules or config? rules is very ambiguous and it does not give any idea as to what is under this package tree.

rshriram (changes requested):

cmd/k8srules/controller.go
```diff
@@ -0,0 +1,273 @@
...
+					{Name: kuberules.ResourceVersion},
+				},
+				Description: kuberules.ResourceDescription,
+			}
+
+			result, err := client.Extensions().ThirdPartyResources().Create(tpr)
+			if err != nil {
+				return err
+			}
+			log.Printf("The ThridPartYresource %#v created\n", result)
```
ijsnellf: Typo: "ThirdPartyResource"

cmd/k8srules/controller.go
```diff
@@ -0,0 +1,273 @@
...
+		} else {
+			return err
+		}
+	} else {
+		log.Printf("The ThirdPartyResource %s already exists\n", resName)
```
ijsnellf: Should we be using logrus instead of the standard logging library? Or alternately, maybe we should put `registry/utils/logging/logging.go` into `/pkg/logging/logging.go` and use that?

ijsnellf (commented):
Few minor comments.

cmd/k8srules/controller.go
zachidan: @rshriram Each rule object is stored as a separated TPR. The sidecar watches the rules endpoint and receives only the rule instance that was changed (that is, the delta).

cmd/k8srules/controller.go
zachidan: @rshriram Only the rule instance that was deleted. Not the entire array of rules.

Makefile
elevran: I would request that we don't run multiple independent and unrelated code in the same process, controlled via command line switches. Let's keep them as separate processes. I can see how it might complicate docs and docker image generation, but taking this argument to its (unreasonable) conclusion implies that we should have all functions (controller, registry, proxy, ...) in a single image and allow selection via invocation flags.

cmd/k8srules/controller.go
elevran: What is the specific issue with importing a8 packages?

cmd/k8srules/controller.go
zachidan: fixed

pkg/adapters/rules/kubernetes/rule.go
zachidan: done

cmd/k8srules/controller.go
zachidan: The controller was modified to use logrus.

@zcahana will move the logging utility from the registry to /pkg/logging

Makefile
rshriram: The reason was that the controller would behave just like the registry and sidecar, With adapter backends for a8api, k8sapi, consul API, etc.

rshriram (approved):
I am hoping that you will add a PR for integration tests as well that uses this new image, after the sidecar starts using the new endpoint

PR #449, opened 2016-12-20
Removed logstash from docker yaml for gremlin

JacobFrericks:
The change to remove logstash was merged right before this was, causing this bug


GregHanson (approved):
LGTM

PR #451, opened 2016-12-22
Added rules adapter for kubernetes

tomerGolany:


pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+	// namespace from which to sync rules
+	namespace string
+
+	// stopChan for stop signals
+	stopChan chan struct{}
+
+	// revision . We assume that sidecar and rules fetcher runs always together
+	revision int
+
+	//ruleChan chan struct{}
```
zcahana: Can remove commented-out code

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+	// mutex is used to synchronize access to the 'services' map,
```
zcahana: Copy-paste error on comment: ('services' map)

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+
+	adapter := &Adapter{
+		client:    client,
+		workqueue: workqueue,
+		rules:     make(map[string]*a8api.Rule),
+		namespace: namespace,
+		revision: 0,
+	}
+	adapter.rulesCache, adapter.rulesController = cache.NewInformer(
+		cache.NewListWatchFromClient(client, fmt.Sprintf("%ss", strings.Replace(ResourceName, "-", "", -1)), namespace, nil),
```
zcahana, line 118: It would be nice to have a code comment on how the resource name is constructed (`fmt.Sprintf("%ss", strings.Replace(ResourceName, "-", "", -1))`).

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+		&RoutingRule{},
+		RulesCacheResyncPeriod,
+		cache.ResourceEventHandlerFuncs{
+			AddFunc:    workqueue.EnqueueingAddFunc(adapter.addRule),
+			UpdateFunc: workqueue.EnqueueingUpdateFunc(adapter.updateRules),
+			DeleteFunc: workqueue.EnqueueingDeleteFunc(adapter.deleteRules),
+		},
+	)
+	logger.Info("Calling Adapter start" )
```
zcahana: This log message is unnecessary (especially on Info level). Please remove.

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+	// mutex is used to synchronize access to the 'services' map,
+	// which is read by the ListXXX() methods (externally),
+	// and is written by the cache event handlers (internally).
+	// Given that we expect a single reader only, we use a regular sync.Mutex rather than a sync.RWMutex.
+	mutex sync.Mutex
+}
+
+// New creates and starts a new Kubernetes Rules adapter.
+func New(config Config) (*Adapter, error) {
+	logger.Info("Adapter starting..." )
```
zcahana: This log message is unnecessary (especially on Info level). Please remove.

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+		&RoutingRule{},
+		RulesCacheResyncPeriod,
+		cache.ResourceEventHandlerFuncs{
+			AddFunc:    workqueue.EnqueueingAddFunc(adapter.addRule),
+			UpdateFunc: workqueue.EnqueueingUpdateFunc(adapter.updateRules),
+			DeleteFunc: workqueue.EnqueueingDeleteFunc(adapter.deleteRules),
+		},
+	)
+	logger.Info("Calling Adapter start" )
+	adapter.workqueue.Start()
```
zcahana: It makes more sense to call `workqueue.Start()` from within the `Adapter.Start()` method (similarly to the Kubernetes discovery adapter). This allows the adapter implementation to be started and stopped more than a single time.

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+				&api.ListOptions{},
+				&api.DeleteOptions{},
+			)
+			return nil
+		})
+	schemeBuilder.AddToScheme(api.Scheme)
+
+	client, err := rest.RESTClientFor(kubeConfig)
+	if err != nil {
+		logger.Printf("Failed to create the in-cluster client. %s\n", err)
```
zcahana: We use `logger.WithError(err).Errorf(...)` for logging errors.

Also note that the log message is not accurate - the client is not necessarily in-cluster...

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+		logger.WithError(err).Errorf("Failed starting Kubernetes adapter")
+		return err
+	}
+	a.stopChan = make(chan struct{})
+
+	go a.rulesController.Run(a.stopChan)
+	return nil
+}
+
+func createRulesClient(config Config) (*rest.RESTClient, error) {
```
zcahana: This should probably re-use the existing `kubepkg.NewTPRClient()` function.
See [here](https://github.com/amalgam8/amalgam8/blob/master/cmd/k8srules/controller.go#L75) for example (but need to pass-in the url/token if present).

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+	defer a.mutex.Unlock()
+
+	rules := make([]a8api.Rule, 0, len(a.rules))
+	for _, rule := range a.rules {
+		rules = append(rules, *rule)
+	}
+	rules = f.Apply(rules)
+	rulesSet := &a8api.RulesSet{
+		Rules:    rules,
+		Revision: 1, // TODO: how to fill revision ?
```
zcahana: Should use `Revision: a.revision`, and make sure `a.revision` is inc'ed with every change to it.

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+
+	}
+	logger.Info("List rules: %v", rulesSet.Rules )
```
zcahana: Remove log statement.

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+	if !ok {
+		logger.Warnf("Invalid rule update: old object is of type %T\n", oldObj)
+		return
+	}
+	newRule, ok := newObj.(*RoutingRule)
+	if !ok {
+		logger.Warnf("Invalid rule update: new object is of type %T\n", newObj)
+		return
+	}
+
```
zcahana: Can avoid processing if resource version hasn't changed (e.g., due to a full resync):
```go
// If resource version hasn't changed (i.e., on full resync), ignore
if oldRule.ResourceVersion >= newRule.ResourceVersion {
	return
}
```

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+		ruleName := newRule.Metadata.Name
+		a.rules[ruleName] = a8Rule
+		return
+	} else {
+		logger.Warnf("Updated rule state %s is undefiend . Rule %#v will not be updated", newRule.Status.State, oldRule)
+		return
+	}
+}
+
+func (a *Adapter) deleteRules(obj interface{}) {
```
zcahana: Note that `obj` here might not be the actual `RoutingRule` object, but a wrapper around it.
There's an `extractDeletedObject()` function in the [Kubernetes discovery adapter](https://github.com/amalgam8/amalgam8/blob/master/pkg/adapters/discovery/kubernetes/kubernetes.go#L495) that handles this; Let's move this function to the `amalgam8/pkg/kubernetes` package, and re-use it from here.

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+// addRule is the callback invoked by the Kubernetes cache when a rule API resource is added.
+func (a *Adapter) addRule(obj interface{}) {
+	rule, ok := obj.(*RoutingRule)
+	if !ok {
+		logger.Warnf("Invalid rule added: object is of type %T", obj)
+		return
+	}
+	if rule.Status.State == RuleStateInvalid {
```
zcahana: Instead of duplicating the logic across `addRule()` and `updateRules()`, add some `storeRule()` method and call it from both. Also note that switching on `rule.Status.State` may be a better fit here.

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,284 @@
...
+		logger.Warnf("Rule state is invalid. Rule %#v will not be added", rule)
+		return
+	} else if rule.Status.State == RuleStateValid {
+		a.revision++
+		a8Rule := &rule.Spec
+		ruleName := rule.Metadata.Name
+		a.rules[ruleName] = a8Rule
+		return
+	} else {
+		logger.Warnf("Rule state %s is undefiend . Rule %#v will not be added", rule.Status.State, rule)
```
zcahana: Type: "undefiend" --> "undefined". Also note that printing "%#v" of the rule is probably too verbose for a log message. Using the rule name will do the trick.

sidecar/sidecar.go
```diff
@@ -284,7 +289,7 @@ func startProxy(conf *config.Config, discovery api.ServiceDiscovery, kubeClient
 	)
 	nginxProxy := proxy.NewNGINXProxy(nginxManager)
 
-	rules, err := buildServiceRules(conf, kubeClient)
```
zcahana, line 273: Can remove the `kubeClient` argument from `startProxy()`. It was there only for the kubernetes rules adapter.

zcahana (changes requested):
Please see inline comments.

Additionally, need to fix code formatting (see Travis build failure) and resolve merge conflicts.

pkg/adapters/rules/kubernetes/kubernetes.go
zcahana: Cancel that. Apparently the workqueue already does that for you.

pkg/adapters/rules/kubernetes/kubernetes.go
```diff
@@ -214,51 +214,34 @@ func (a *Adapter) ListRules(f *a8api.RuleFilter) (*a8api.RulesSet, error) {
...
-		return
-	}
-	newRule, ok := newObj.(*RoutingRule)
+	a.storeRules(&newObj)
+
+}
+
+// storeRules is a helper function called by updateRules() and addRule() callback functions. is stores or updates rules
+// int the rules map of the adapter.
+func (a *Adapter) storeRules(obj *interface{}) {
```
zcahana: `*interface{}` type is confusing and unnecessary. Please declare the function argument to be `obj interface{}`. and invoke it with `a.storeRules(obj)` and _not_ `a.storeRules(&obj)`.

pkg/adapters/rules/kubernetes/kubernetes.go
zcahana: Still open, please address.

sidecar/sidecar.go
zcahana: Still open, please address.

zcahana:
@tomerGolany Please go through the prev code review comments - there are 2 items that are still open. Additionally, see an additional minor comment on new commits. Otherwise, this looks fine and can be merged.

zcahana (approved):

PR #463, opened 2017-01-09
Gremlin integration testing with new CLI

gicaceres:
The purpose of these changes is to include Gremlin in the integration test using the new CLI.

- Updated filebeat from 1.2.2 to 5.1.1 in the sidecard
- Run filebeat in the productpage and reviews apps
- Added Elasticsearch and Gremlin to the controlplane
- Consolidated controlplane files for docker and k8
- Replace old CLI (python) with new CLI (Go) in the integration tests
- Added Gremlin recipe test in the integration test


testing/docker/run-controlplane-docker.sh
```diff
@@ -20,6 +20,9 @@ set -x
 SCRIPTDIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 if [ "$1" == "start" ]; then
+    # Increase memory limit for elasticsearch 5.1
+    sudo sysctl -w vm.max_map_count=262144
+
     echo "starting Control plane components (registry, and controller)"
     docker-compose -f $SCRIPTDIR/controlplane.yaml up -d
```
rshriram: Do we even need this file, as it is no longer doing anything special. I suggest we simply add the following two lines to test-docker.sh
```bash
sysctl -w vm.max_map_count=...
docker-compose -f $SCRIPTDIR/controlplane.yaml up -d
```

Similarly, we can get rid of cleanup.sh and move the two lines into that file. Given that the chances of errors in starting the control plane services is low in travis, we can do away with all the tests (loop, check for health, etc.). May be have it commented. WDYT?

testing/docker/test-docker.sh
```diff
@@ -24,9 +24,6 @@ echo "Testing docker-based deployment.."
 $SCRIPTDIR/run-controlplane-docker.sh start
 sleep 5
 
-docker-compose -f $SCRIPTDIR/gateway.yaml up -d
-sleep 5
-
 docker-compose -f $SCRIPTDIR/bookinfo.yaml up -d
 sleep 10
```
rshriram, line 35: please refer to my comments above. Cleaning up these files is long overdue.

testing/kubernetes/cleanup.sh
```diff
@@ -17,8 +17,6 @@
 set -x
 SCRIPTDIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
-kubectl delete -f $SCRIPTDIR/gateway.yaml
-sleep 2
 kubectl delete -f $SCRIPTDIR/bookinfo.yaml
 sleep 2
 $SCRIPTDIR/run-controlplane-kubernetes.sh stop
```
rshriram: Same comment as the docker case. We can fold these two files cleanup.sh and run-controlplane-kubernetes.sh

testing/kubernetes/test-kubernetes.sh
```diff
@@ -27,9 +27,6 @@ sleep 10
 $SCRIPTDIR/run-controlplane-kubernetes.sh start
 sleep 10
 
-kubectl create -f $SCRIPTDIR/gateway.yaml
-sleep 5
-
 kubectl create -f $SCRIPTDIR/bookinfo.yaml
 echo "Waiting for the services to come online.."
 sleep 10
```
rshriram, line 36: same comment as docker

testing/kubernetes/run-controlplane-kubernetes.sh
```diff
@@ -54,9 +52,11 @@ if [ "$1" == "start" ]; then
...
-    kubectl delete -f $SCRIPTDIR/$cfile
+    kubectl delete -f $SCRIPTDIR/$ctrlplanefile
     sleep 3
-    kubectl delete -f $SCRIPTDIR/$rdsfile
+    # kubectl delete -f $SCRIPTDIR/$cfile
+    # sleep 3
+    # kubectl delete -f $SCRIPTDIR/$rdsfile
 else
     echo "usage: $0 start|stop"
     exit 1
```
rshriram: kill file

testing/test-scripts/demo_script.sh
```diff
@@ -14,30 +14,74 @@
...
-    
+
     # Diff, but ignore all whitespace since
-    diff -Ewb <(echo $file1) <(echo $file2)        
+    diff -Ewb <(echo $file1) <(echo $file2)
 }
 
 SCRIPTDIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+make -C $SCRIPTDIR/../../ build.cli
+CLIBIN=$SCRIPTDIR/../../bin/a8ctl-beta
```
rshriram: Do we need to compile cli here? Isn't it already compiled before reaching these test scripts?
If its not, I think its time that we promote the CLI to be part of the main build (may be not release but atleast main build), at Travis level.

testing/test-scripts/demo_script.sh
```diff
@@ -14,30 +14,74 @@
...
+    - tags:
+      - version=v2
+- priority: 1
+  destination: reviews
+  route:
+    backends:
+    - tags:
+      - version=v1
+EOF
 sleep 10
```
rshriram, line 49: I suggest moving these to separate YAML files, to ease maintenance in future.

testing/test-scripts/demo_script.sh
```diff
@@ -133,3 +197,29 @@ if [ $? -gt 0 ]; then
     exit 1
 fi
 echo "works!"
+
+#######Gremlin
+if [ "$A8_TEST_GREMLIN" = true ]; then
+  GREMLIN_FILES=$SCRIPTDIR/../../examples
+  echo "Testing gremlin recipe.."
+  echo "please wait.."
```
rshriram, line 151: remove this line

rshriram (changes requested):
Minor changes related to cleaning up old files. Otherwise, it looks good.

testing/test-scripts/demo_script.sh
rshriram: @zcahana what do you think?

testing/docker/run-controlplane-docker.sh
gicaceres: Sure, I think we can use test-docker.sh and remove the other files.

testing/test-scripts/demo_script.sh
gicaceres: We can build the CLI before this line:
https://github.com/amalgam8/amalgam8/blob/master/.travis.yml#L40

gicaceres: On a second thought, I think it would be better if we build the CLI in the makefile (as a dependency of test.integration) otherwise users will have to build the CLI manually before running the integ test locally.

testing/docker/test-docker.sh
gicaceres: done

testing/kubernetes/test-kubernetes.sh
gicaceres: done

testing/test-scripts/demo_script.sh
gicaceres: done

testing/kubernetes/run-controlplane-kubernetes.sh
gicaceres: done

testing/kubernetes/cleanup.sh
gicaceres: done

testing/test-scripts/demo_script.sh
rshriram: With the changes to makefile #464 , we no longer need to build the cli here. Just set
```bash
CLIBIN=${SCRIPTDIR}/../../bin/a8ctl-beta-linux
```

rshriram (approved):

PR #466, opened 2017-01-11
Fix regex used in cli and tests

rshriram:
Fix regex such that it is compatible with envoy and nginx. It also fixes buggy regex (earlier one matched foouser=jason and user=jason


GregHanson (approved):
LGTM

PR #467, opened 2017-01-11
Add makefile target to build the example apps.

craigyam:
Update the makefile to add the build.exampleapps target that will
build the example helloworld and bookinfo apps.


rshriram (approved):
Sorry for delay. Can you change this to dockerize.exampleapps? With that the pr is good

PR #485, opened 2017-01-25
Change the service encode/decode logic.

ijsnellf:
Removed support for escaped delimiters and switched to the standard format for encoding/decoding service name and tags:

```service:tag1=value1,tag2=value2```


GregHanson (approved):
LGTM

rshriram (approved):

Issue #476, opened 2017-01-23
Rules DSL: support circuit breaker configuration for Envoy


rshriram:

rshriram:
fixed by #492

PR #492, opened 2017-01-30
adding circuit breakers to envoy

GregHanson:
Addresses:
https://github.com/amalgam8/amalgam8/issues/476

### DIFF
--- a/pkg/api/rule.go
+++ b/pkg/api/rule.go
@@ -41,7 +41,9 @@ type Match struct {
 
 // Route definition
 type Route struct {
-	Backends []Backend `json:"backends"`
+	Backends          []Backend `json:"backends"`
+	TCPConnectTimeout float64   `json:"tcp_connect_timeout,omitempty"`
+	MaxRetries        int       `json:"max_retries,omitempty"`
 }
 
 // URI for backends.
@@ -51,25 +53,24 @@ type URI struct {
 	PrefixRewrite string `json:"prefix_rewrite"`
 }
 
-type Hystrix struct {
-	TimeoutMS                float64 `json:"timeout_ms,omitempty"`
-	MaxRetries               int     `json:"max_retries,omitempty"`
-	MaxConnections           int     `json:"max_connections,omitempty"`
-	MaxPendingRequest        int     `json:"max_pending_requests,omitempty"`
-	MaxRequests              int     `json:"max_requests,omitempty"`
-	SleepWindowMS            int     `json:"sleep_window_ms,omitempty"`
-	ConsecutiveErrors        int     `json:"consecutive_errors,omitempty"`
-	DetectionIntervalMS      int     `json:"detection_interval_ms,omitempty"`
-	MaxRequestsPerConnection int     `json:"max_requests_per_connection,omitempty"`
+// Resilience cluster level circuit breaker options
+type Resilience struct {
+	MaxConnections           int `json:"max_connections,omitempty"`
+	MaxPendingRequest        int `json:"max_pending_requests,omitempty"`
+	MaxRequests              int `json:"max_requests,omitempty"`
+	SleepWindow              int `json:"sleep_window,omitempty"`
+	ConsecutiveErrors        int `json:"consecutive_errors,omitempty"`
+	DetectionInterval        int `json:"detection_interval,omitempty"`
+	MaxRequestsPerConnection int `json:"max_requests_per_connection,omitempty"`
 }
 
 // Backend represents a backend to route to.
 type Backend struct {
-	Name    string   `json:"name,omitempty"`
-	Tags    []string `json:"tags"`
-	URI     *URI     `json:"uri,omitempty"`
-	Weight  float64  `json:"weight,omitempty"`
-	Hystrix *Hystrix `json:"hystrix,omitempty"`
+	Name       string      `json:"name,omitempty"`
+	Tags       []string    `json:"tags"`
+	URI        *URI        `json:"uri,omitempty"`
+	Weight     float64     `json:"weight,omitempty"`
+	Resilience *Resilience `json:"resilience,omitempty"`
 }
 
 // Action to take.
--- a/pkg/api/rule.go
+++ b/pkg/api/rule.go
@@ -41,9 +41,9 @@ type Match struct {
 
 // Route definition
 type Route struct {
-	Backends          []Backend `json:"backends"`
-	TCPConnectTimeout float64   `json:"tcp_connect_timeout,omitempty"`
-	MaxRetries        int       `json:"max_retries,omitempty"`
+	Backends       []Backend `json:"backends"`
+	HTTPReqTimeout float64   `json:"http_req_timeout,omitempty"`
+	HTTPReqRetries int       `json:"http_req_retries,omitempty"`
 }
 
 // URI for backends.
@@ -55,13 +55,13 @@ type URI struct {
 
 // Resilience cluster level circuit breaker options
 type Resilience struct {
-	MaxConnections           int `json:"max_connections,omitempty"`
-	MaxPendingRequest        int `json:"max_pending_requests,omitempty"`
-	MaxRequests              int `json:"max_requests,omitempty"`
-	SleepWindow              int `json:"sleep_window,omitempty"`
-	ConsecutiveErrors        int `json:"consecutive_errors,omitempty"`
-	DetectionInterval        int `json:"detection_interval,omitempty"`
-	MaxRequestsPerConnection int `json:"max_requests_per_connection,omitempty"`
+	MaxConnections           int     `json:"max_connections,omitempty"`
+	MaxPendingRequest        int     `json:"max_pending_requests,omitempty"`
+	MaxRequests              int     `json:"max_requests,omitempty"`
+	SleepWindow              float64 `json:"sleep_window,omitempty"`
+	ConsecutiveErrors        int     `json:"consecutive_errors,omitempty"`
+	DetectionInterval        float64 `json:"detection_interval,omitempty"`
+	MaxRequestsPerConnection int     `json:"max_requests_per_connection,omitempty"`
 }
 
 // Backend represents a backend to route to.
--- a/pkg/api/rule.go
+++ b/pkg/api/rule.go
@@ -51,14 +51,25 @@ type URI struct {
 	PrefixRewrite string `json:"prefix_rewrite"`
 }
 
+type Hystrix struct {
+	TimeoutMS                float64 `json:"timeout_ms,omitempty"`
+	MaxRetries               int     `json:"max_retries,omitempty"`
+	MaxConnections           int     `json:"max_connections,omitempty"`
+	MaxPendingRequest        int     `json:"max_pending_requests,omitempty"`
+	MaxRequests              int     `json:"max_requests,omitempty"`
+	SleepWindowMS            int     `json:"sleep_window_ms,omitempty"`
+	ConsecutiveErrors        int     `json:"consecutive_errors,omitempty"`
+	DetectionIntervalMS      int     `json:"detection_interval_ms,omitempty"`
+	MaxRequestsPerConnection int     `json:"max_requests_per_connection,omitempty"`
+}
+
 // Backend represents a backend to route to.
 type Backend struct {
 	Name    string   `json:"name,omitempty"`
 	Tags    []string `json:"tags"`
 	URI     *URI     `json:"uri,omitempty"`
 	Weight  float64  `json:"weight,omitempty"`
-	Timeout float64  `json:"timeout,omitempty"`
-	Retries int      `json:"retries,omitempty"`
+	Hystrix *Hystrix `json:"hystrix,omitempty"`
 }
 
 // Action to take.
--- a/pkg/api/rules-schema.json
+++ b/pkg/api/rules-schema.json
@@ -165,17 +165,65 @@
           "exclusiveMinimum": true,
           "default": 1
         },
-        "timeout": {
-          "description": "Connect/read/write timeout when talking to an instance. Default is nginx's timeout (60s). Unit is in seconds (e.g, 1.234s)",
-          "type": "number",
-          "minimum": 0,
-          "exclusiveMinimum": true
-        },
-        "retries": {
-          "description": "Number of times to retry a failed request to a different instances. Set to 0 to disable retries",
-          "type": "number",
-          "minimum": 0,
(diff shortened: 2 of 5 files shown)


sidecar/proxy/envoy/manager.go
```diff
@@ -312,6 +313,20 @@ func buildClusters(rules []api.Rule) []Cluster {
 			ConnectTimeoutMs: 1000,
 		}
 
+		if backend.Timeout > 0 {
+			// convert from float to int
+			cluster.ConnectTimeoutMs = int(backend.Timeout * 1000)
+		}
```
rshriram: To avoid this format conversion, I suggest we change the rules schema to use timeout_budget_ms. What do you think?

GregHanson:
@rshriram `timeout` and `retries` are in the rules schema under the `Backend`s field.  Should I add the remaining circuit breaker fields:

```
	MaxConnections    int `json:"max_connections,omitempty"`
	MaxPendingRequest int `json:"max_pending_requests,omitempty"`
	MaxRequests       int `json:"max_requests,omitempty"`
```

under the Backend struct as well?  Or should I make a substruct in Backend?  Not sure if adding the substruct in the rule schema will break any retry/timeout logic in the nginx lua code

sidecar/proxy/envoy/manager.go
```diff
@@ -312,6 +313,20 @@ func buildClusters(rules []api.Rule) []Cluster {
...
+
+		if backend.Retries > 0 {
+			cluster.Hystrix = CircuitBreaker{
+				MaxRetries:        backend.Retries,
+				MaxConnections:    50,
+				MaxRequests:       25,
+				MaxPendingRequest: 5,
+			}
```
rshriram: hard coding?

rshriram (changes requested):

sidecar/proxy/envoy/manager.go
ijsnellf: See @GregHanson's comment above.

sidecar/proxy/envoy/manager.go
ijsnellf: That will mean we have to do the inverse conversion in NGINX. Do we care about keeping NGINX up to date?

rshriram: no..

sidecar/proxy/envoy/manager.go
```diff
@@ -292,24 +292,62 @@ func ParseServiceKey(s string) (string, []string) {
...
+			Hystrix:          CircuitBreaker{},
+			OutlierDetection: OutlierDetection{
+				MaxEjectionPercent: 100,
+			},
+		}
+
+		if backend.Hystrix != nil {
+			if backend.Hystrix.TimeoutMS > 0 {
+				// convert from float sec to int ms
+				cluster.ConnectTimeoutMs = int(backend.Hystrix.TimeoutMS * 1000)
```
ijsnellf: It is confusing that the field is called `timeout_ms`, but apparently contains float seconds. Either we should switch to a field name of just `timeout`, or we should have `timeout_ms` contain integer milliseconds.

pkg/api/rules-schema.json
```diff
@@ -165,17 +165,65 @@
...
-        "retries": {
-          "description": "Number of times to retry a failed request to a different instances. Set to 0 to disable retries",
-          "type": "number",
-          "minimum": 0,
-          "exclusiveMinimum": false
+        "hystrix": {
+          "type": "object",
+          "properties": {
+            "timeout_ms": {
+              "description": "Connect/read/write timeout when talking to an instance. Default is nginx's timeout (60s). Unit is in seconds (e.g, 1.234s)",
```
ijsnellf: Remove reference to NGINX.

pkg/api/rules-schema.json
```diff
@@ -165,17 +165,65 @@
...
+              "type": "number",
+              "minimum": 0,
+              "exclusiveMinimum": true
+            },
+            "max_retries": {
+              "description": "Maximum number of retries to connect to a backend",
```
ijsnellf: Mention defaults?

ijsnellf: Same comment for everything below...

sidecar/proxy/envoy/resources.go
```diff
@@ -134,13 +135,29 @@ type Host struct {
...
-	MaxRequestsPerConnection int    `json:"max_requests_per_connection,omitempty"`
-	Hosts                    []Host `json:"hosts,omitempty"`
+	Name                     string           `json:"name"`
+	ServiceName              string           `json:"service_name,omitempty"`
+	ConnectTimeoutMs         int              `json:"connect_timeout_ms"`
+	Type                     string           `json:"type"`
+	LbType                   string           `json:"lb_type"`
+	MaxRequestsPerConnection int              `json:"max_requests_per_connection,omitempty"`
+	Hosts                    []Host           `json:"hosts,omitempty"`
+	Hystrix                  CircuitBreaker   `json:"circuit_breaker"`
```
ijsnellf: Field name should match the JSON tag to reduce confusion.

sidecar/proxy/envoy/resources.go
```diff
@@ -70,6 +70,7 @@ type Route struct {
 	PrefixRewrite string   `json:"prefix_rewrite,omitempty"`
 	Cluster       string   `json:"cluster"`
 	Headers       []Header `json:"headers,omitempty"`
+	Timeout       string   `json:"timeout_ms,omitempty"`
```
ijsnellf: Was this added accidentally? I don't see it used anywhere, and it is type `string`.

sidecar/proxy/envoy/manager.go
```diff
@@ -292,24 +292,62 @@ func ParseServiceKey(s string) (string, []string) {
...
 			Type:             "sds",
 			LbType:           "round_robin",
 			ConnectTimeoutMs: 1000,
+			Hystrix:          CircuitBreaker{},
+			OutlierDetection: OutlierDetection{
+				MaxEjectionPercent: 100,
+			},
+		}
+
+		if backend.Hystrix != nil {
```
ijsnellf: Put this in its own function?

sidecar/proxy/envoy/manager.go
```diff
@@ -292,24 +292,62 @@ func ParseServiceKey(s string) (string, []string) {
 }
 
 func buildClusters(rules []api.Rule) []Cluster {
-	clusterMap := make(map[string]struct{})
+	clusterMap := make(map[string]*api.Backend)
```
ijsnellf, line 295: What about clusters with more than one backend?

ijsnellf (changes requested):

sidecar/proxy/envoy/manager.go
rshriram: I agree. We should change the json schema spec to timeout_ms and use timeout ms everywhere.

pkg/api/rules-schema.json
rshriram: and say its in milliseconds

sidecar/proxy/envoy/manager.go
GregHanson: @rshriram changing this mean changing all of our docs for bookinfo where we tell everyone to create a 7 second delay - still want me to go forward with that?

rshriram: On second thought, I take it back. Lets revert to using consistent "timeout".. and eliminate all the _ms and MS stuff. In an future PR, we can revisit changing all these to milliseconds.

GregHanson:
Here is what a sample rule will look like with the new schema:

```
    {
      "id": "1e161ec1-b2ce-4518-a031-29ee56645918",
      "priority": 1,
      "destination": "productpage",
      "route": {
        "backends": [
          {
            "tags": [
              "version=v1"
            ],
            "resilience": {
              "max_connections": 4,
              "max_requests": 88,
              "consecutive_errors": 9,
              "max_requests_per_connection": 1021
            }
          }
        ],
        "tcp_connect_timeout": 0.5,
        "max_retries": 14
      }
    }
```

GregHanson:
Here is what the sample envoy.json file will look like:

```
{
  "runtime": {
    "symlink_root": "/etc/envoy/runtime/routing",
    "subdirectory": "traffic_shift"
  },
  "listeners": [
    {
      "port": 6379,
      "filters": [
        {
          "type": "read",
          "name": "http_connection_manager",
          "config": {
            "codec_type": "auto",
            "stat_prefix": "ingress_http",
            "generate_request_id": true,
            "add_user_agent": true,
            "route_config": {
              "virtual_hosts": [
                {
                  "name": "backend",
                  "domains": [
                    "*"
                  ],
                  "routes": [
                    {
                      "runtime": {
                        "key": "reviews._:version=v1",
                        "default": 0
                      },
                      "prefix": "/reviews/",
                      "prefix_rewrite": "/",
                      "cluster": "reviews:version=v1",
                      "retry_policy": {
                        "retry_on": "5xx,connect-failure,refused-stream"
                      }
                    },
                    {
                      "runtime": {
                        "key": "productpage._:version=v1",
                        "default": 0
                      },
                      "prefix": "/productpage/",
                      "prefix_rewrite": "/",
                      "cluster": "productpage:version=v1",
                      "timeout_ms": 500,
                      "retry_policy": {
                        "retry_on": "5xx,connect-failure,refused-stream",
                        "num_retries": 14
                      }
                    }
              ]
            },
            "filters": [
              {
                "type": "decoder",
                "name": "router",
                "config": {
                  "dynamic_stats": false
                }
              }
            ],
            "access_log": [
              {
                "path": "/var/log/a8_access.log",
                "format": "{\"status\":\"%RESPONSE_CODE%\", \"start_time\":\"%START_TIME%\", \"request_time\":%DURATION%, \"upstream_response_time\":\"%RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)%\", \"src\":\"gateway:version=none\", \"dst\":\"%UPSTREAM_CLUSTER%\", \"gremlin_recipe_id\":\"-\", \"message\":\"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %RESPONSE_CODE%\", \"module\":\"ENVOY\", \"method\":\"%REQ(:METHOD)%\", \"path\":\"%REQ(X-ENVOY-ORIGINAL-PATH?:PATH)%\", \"protocol\":\"%PROTOCOL%\", \"response_flags\":\"%RESPONSE_FLAGS%\", \"x_forwarded\":\"%REQ(X-FORWARDED-FOR)%\", \"user_agent\":\"%REQ(USER-AGENT)%\", \"request_id\":\"%REQ(X-REQUEST-ID)%\", \"auth\":\"%REQ(:AUTHORITY)%\", \"upstream_host\":\"%UPSTREAM_HOST%\"}\n"
              }
            ]
          }
        }
      ]
    }
  ],
  "admin": {
    "access_log_path": "/var/log/envoy_admin.log",
    "port": 8001
  },
  "cluster_manager": {
    "clusters": [
      {
        "name": "details:version=v1",
        "service_name": "details:version=v1",
        "connect_timeout_ms": 1000,
        "type": "sds",
        "lb_type": "round_robin",
        "circuit_breaker": {},
        "outlier_detection": {
          "max_ejection_percent": 100
        }
      },
      {
        "name": "productpage:version=v1",
        "service_name": "productpage:version=v1",
        "connect_timeout_ms": 1000,
        "type": "sds",
        "lb_type": "round_robin",
        "max_requests_per_connection": 1021,
        "circuit_breaker": {
          "max_connections": 4,
          "max_requests": 88
        },
        "outlier_detection": {
          "consecutive_5xx": 9,
          "max_ejection_percent": 100
        }
      }
    ],
    "sds": {
      "cluster": {
        "name": "sds",
        "connect_timeout_ms": 1000,
        "type": "strict_dns",
        "lb_type": "round_robin",
        "max_requests_per_connection": 1,
        "hosts": [
          {
            "url": "tcp://127.0.0.1:6500"
          }
        ]
      },
      "refresh_delay_ms": 1000
    }
  }
}
```

Note that productpage `route` and `cluster` have the fields configured while the other services demonstrate the default settings

pkg/api/rule.go
```diff
@@ -41,7 +41,9 @@ type Match struct {
 
 // Route definition
 type Route struct {
-	Backends []Backend `json:"backends"`
+	Backends          []Backend `json:"backends"`
+	TCPConnectTimeout float64   `json:"tcp_connect_timeout,omitempty"`
```
rshriram: HttpReqTimeout

pkg/api/rule.go
```diff
@@ -41,7 +41,9 @@ type Match struct {
...
+	MaxRetries        int       `json:"max_retries,omitempty"`
 }
```
rshriram: HttpReqRetries

pkg/api/rules-schema.json
```diff
@@ -192,6 +228,18 @@
           "items": {
             "$ref": "#/definitions/backend"
           }
+        },
+        "tcp_connect_timeout": {
+          "description": "Connect/read/write timeout when talking to a route. Default is 15. Unit is in seconds.",
```
rshriram: http_req_timeout
Timeout for a HTTP request. Includes retries as well. Unit is in floating point seconds. Default 15.0s

pkg/api/rules-schema.json
```diff
@@ -192,6 +228,18 @@
...
+          "type": "number",
+          "minimum": 0,
+          "exclusiveMinimum": true
+        },
+        "max_retries": {
+          "description": "Maximum number of retries to connect to a route.  Defaults to 1.",
```
rshriram: http_req_retries
Number of retries for a given request.  The interval between retries will be determined automatically (25ms+). Actual number of retries attempted depends on the http_req_timeout

rshriram (changes requested):

rshriram (approved):

sidecar/proxy/envoy/manager.go
```diff
@@ -292,24 +292,60 @@ func ParseServiceKey(s string) (string, []string) {
...
 			Type:             "sds",
 			LbType:           "round_robin",
 			ConnectTimeoutMs: 1000,
+			CircuitBreaker:   &CircuitBreaker{},
+			OutlierDetection: &OutlierDetection{
+				MaxEjectionPercent: 100,
+			},
+		}
+
+		if backend.Resilience != nil {
```
ijsnellf: No function? That's ok. That's fine.

PR #494, opened 2017-01-31
Load Balancing Algorithm Options Added

kimikowang:


pkg/api/rule.go
```diff
@@ -59,6 +59,7 @@ type Backend struct {
 	Weight  float64  `json:"weight,omitempty"`
 	Timeout float64  `json:"timeout,omitempty"`
 	Retries int      `json:"retries,omitempty"`
+	Lb_type string 	 `json:"lb_type,omitempty"`
```
ijsnellf: Variable naming style:
```
LbType string 	 `json:"lb_type,omitempty"`
```

rshriram (changes requested):
You still have to pass this to envoy config right?
Can you verify that the generated config file for envoy actually has the right load balancing algorithm?

rshriram (changes requested):
Please remove productpage binary and the cli/i18n/resources.go. Otherwise it looks good.

kimikowang:
Changes have been made to sidecar/proxy/envoy/manager.go, where LbType value in backend rule will be included in cluster building to generate the config file.

rshriram:
See my comment above. You have included irrelevant files in your commit.

rshriram (changes requested):
And you added another binary.

rshriram (commented):
i18n_resources.go is still there.

rshriram (approved):

PR #496, opened 2017-02-02
tidy up typos and missing targets

elevran:


Makefile
```diff
@@ -305,7 +305,7 @@ release.controller:
 	@mkdir -p $(RELEASEDIR)
 	@tar -czf $(RELEASEDIR)/$(CONTROLLER_RELEASE_NAME).tar.gz --transform 's:^.*/::' $(BINDIR)/$(CONTROLLER_APP_NAME) README.md LICENSE
 
-release.sidecar.envoy:
+release.sidecar.envoy: build.sidecar
```
zcahana: This is inconsistent with the rest of the `release.*` targets, which do not explicitly depend on the corresponding `build.*` targets. Any reason for this one to be different?

zcahana (approved):

Makefile
elevran: not really. they should all be the same, this is just the one I was using. It can still be called explicitly in make, without setting the dependency. Will revert this line

rshriram:
Do we even have yaml files in examples directory?
I think they were moved to the testing directory with variable substitution, so that the same yaml files can be used for both testing go version of bookinfo and the release as well.

elevran:
@rshriram this is a makefile change, I think you meant to comment on a different PR

rshriram:
Yes sorry. It was the previous pr but that was merged. So I thought I would comment on this one, so that you could move the file to the right place.

Issue #490, opened 2017-01-27
Sidecar: Use weighted routing rule in Envoy


rshriram:

jemcginn:
P1 - Need for bookinfo

rshriram:
fixed by #505. We are not going to use weighted_routing for this iteration

PR #505, opened 2017-02-07
simplified weighted cluster generation

GregHanson:
https://github.com/amalgam8/amalgam8/issues/490

### DIFF
--- a/sidecar/proxy/envoy/manager.go
+++ b/sidecar/proxy/envoy/manager.go
@@ -19,7 +19,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"math/rand"
 	"os"
 	"path/filepath"
@@ -175,7 +174,7 @@ func (m *manager) generateConfig(rules []api.Rule, instances []api.ServiceInstan
 	routes := buildRoutes(rules)
 	filters := buildFaults(rules, inst.ServiceName, inst.Tags)
 
-	if err := buildFS(rules, m.workingDir); err != nil {
+	if err := buildFS(m.workingDir); err != nil {
 		return Config{}, err
 	}
 
@@ -405,7 +404,7 @@ func buildRoutes(ruleList []api.Rule) []Route {
 
 				runtime := &Runtime{
 					Key:     buildWeightKey(backend.Name, backend.Tags),
-					Default: 0,
+					Default: int(backend.Weight * 100),
 				}
 
 				route := Route{
@@ -485,6 +484,12 @@ func sanitizeRules(ruleList []api.Rule) {
 					}
 				}
 			}
+
+			sum = 0
+			for j := range rule.Route.Backends {
+				sum += rule.Route.Backends[j].Weight
+				rule.Route.Backends[j].Weight = sum
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


sidecar/proxy/envoy/manager.go
```diff
@@ -405,7 +404,7 @@ func buildRoutes(ruleList []api.Rule) []Route {
 
 				runtime := &Runtime{
 					Key:     buildWeightKey(backend.Name, backend.Tags),
-					Default: 0,
+					Default: int(backend.Weight * 100),
```
rshriram, line 415: Let's change the spec to use integer weights

sidecar/proxy/envoy/manager.go
```diff
@@ -485,6 +484,12 @@ func sanitizeRules(ruleList []api.Rule) {
 					}
 				}
 			}
+
+			sum = 0
+			for j := range rule.Route.Backends {
+				sum += rule.Route.Backends[j].Weight
+				rule.Route.Backends[j].Weight = sum
+			}
```
rshriram, line 500: We don't need this with  weighted_clusters right? The weights there are absolute weights.

sidecar/proxy/envoy/manager.go
```diff
@@ -565,60 +548,6 @@ func buildFS(ruleList []api.Rule, workingDir string) error {
 		return err
 	}
 
-	dirName, err := ioutil.TempDir(workingDir+runtimeVersionsPath, "")
-	if err != nil {
```
rshriram: Isn't this useful for things like fault injection percentages ?

rshriram (changes requested):

sidecar/proxy/envoy/manager.go
GregHanson: If setting those values in the runtime path is configurable it is not documented [here](https://lyft.github.io/envoy/docs/configuration/http_filters/fault_filter.html).  we are not currently doing this for faults

sidecar/proxy/envoy/manager.go
GregHanson: @rshriram what is the benefit of using [Approach A](https://lyft.github.io/envoy/docs/configuration/http_conn_man/route_config/traffic_splitting.html#a-weight-based-cluster-selection) over [Approach B](https://lyft.github.io/envoy/docs/configuration/http_conn_man/route_config/traffic_splitting.html#b-probabilistic-route-selection)?  Right now this PR is still using Approach B

We were originally using approach B because we could edit the filesystem values without needing to do a hot reload of envoy

sidecar/proxy/envoy/manager.go
GregHanson: by spec you mean the rules schema?  Just to verify this will mean re-writing all of the examples docs and CLI to use integers vs floats when creating the rules

rshriram: hmm.. Okay, now that sounds like a bad idea.. Lets stick to this current format.

sidecar/proxy/envoy/manager.go
rshriram: You can edit the filesystem with weighted_clusters as well.
Secondly, in Approach B (probability based cluster selection), it is incorrect to use the term weights. Look at the example in Envoy. To split traffic into 33/33/33, you need to set weight1 to 33, weight2 to 50 (not 66), and nothing for last case.

It just grows infinitely more complicated..

With weighted clusters, its super simple. Each cluster has the same percentage weight as specified by user. And you can adjust the weights by using the runtime_key_prefix.. E.g., 
to adjust weight of cluster1, you tweak the value of runtime_key_prefix+cluster1.

sidecar/proxy/envoy/manager.go
rshriram: It is documented. https://lyft.github.io/envoy/docs/configuration/http_filters/fault_filter.html#runtime. We can get rid of all runtimes if you want to, but then we will have to make sure to leverage Envoy's RDS. And make do with envoy hot reload whenever we tweak fault injection stuff.

rshriram: can you just keep this as a separate function or comment it out, instead of removing the functionality?

rshriram (approved):

PR #507, opened 2017-02-08
K8srule patching issue 502

### LINKED ISSUE #502: k8srules controller overwrites id, destination and priority
Given the following rule definition:
```yaml
apiVersion: "amalgam8.io/v1"
kind: RoutingRule
metadata:
  name: "test-hw"
spec:
  priority: 1
  id: "test"
  destination: "helloworld"
  route:
    backends:
    - tags:
      - "version=v1"
```
When the status is set to `valid`, the `id`, `destination` and `priority` fields are nullified (i.e., set to empty strings and 0, respectively). This is most likely due to the interaction between the resource patching in the controller and not having `omitempty` set for these fields.

elevran:
Do not marshal nil fields from struct to allow correct resource PATCH.
Validated with sample from issue, resulting in:
```yaml
{
    "apiVersion": "v1",
    "items": [
        {
            "apiVersion": "amalgam8.io/v1",
            "kind": "RoutingRule",
            "metadata": {
                "creationTimestamp": "2017-02-08T11:28:00Z",
                "name": "set-helloworld-default-v1",
                "namespace": "default",
                "resourceVersion": "1835274",
                "selfLink": "/apis/amalgam8.io/v1/namespaces/default/routingrules/set-helloworld-default-v1",
                "uid": "a5bffd3e-edf1-11e6-a49c-fa163efdea7d"
            },
            "spec": {
                "destination": "helloworld",
                "id": "set-helloworld-default-v1",
                "priority": 1,
                "route": {
                    "backends": [
                        {
                            "tags": [
                                "version=v1"
                            ]
                        }
                    ]
                }
            },
            "status": {
                "state": "valid"
            }
        }
    ],
    "kind": "List",
    "metadata": {},
    "resourceVersion": "",
    "selfLink": ""
}
```

Fixes #502


rshriram (changes requested):
@elevran the tests are failing. Seems like changing the fields to omit empty is insufficient. You might have to add additional code to the controller to play well with automatic rule id generation.

rshriram:
By failing tests, I mean the docker use cases as well. See Travis logs.

elevran:
Since commit [d037b9b](https://github.com/amalgam8/amalgam8/pull/507/commits/d037b9ba854e1a0e12906724145b8da202f0108b) passes tests - it is not an issue with `omitempty`. 
The controller is rejecting the rules since there is no `id` in them.
The difference between k8s rules and a8 controller is that k8s assigns the `id` if it's missing and then calls validate, whereas a8controller calls validate (failing due to missing id) and only then sets the id.
I will resubmit by making id **not** a required attribute. It is automatically generated.

rshriram:
So, in the k8s case, if user fails to provide an ID, what happens? a random ID is assigned to the rule, just like what we do today in default a8 controller ?

elevran:
k8srules copies the id from the metadata name if missing. By chance (or not) it was passing since we had the object metadata.name set to the rule description...
But it could use also port to etcd and then use that id for the rule.
Best to **not** require `id`. Resubmitted with this change

rshriram (approved):

PR #509, opened 2017-02-08
strict_dns support for cf apps from registry

GregHanson:
finds instances from registry and build a static dns cluster for them in envoy.json file


rshriram:
Hold off on this please. It requires changes at Envoy level as well. This alone won't suffice

rshriram:
You might want to continue this again. Basically if we have CDS, and we return strict dns, this should work with the auto_host_rewrite option set in the route. See Envoy pr

sidecar/proxy/envoy/manager.go
```diff
@@ -330,65 +333,67 @@ func BuildClusters(instances []*api.ServiceInstance, rules []api.Rule, tlsConfig
 	rules = AddDefaultRouteRules(rules, instances)
 
 	backends := make(map[string]*api.Backend)
+	staticClusters := make(map[string]*api.Backend)
+	instanceByClusterMap := make(map[string][]*api.ServiceInstance)
```
ijsnellf: `instancesByClusterMap`?

sidecar/proxy/envoy/manager.go
```diff
@@ -330,65 +333,67 @@ func BuildClusters(instances []*api.ServiceInstance, rules []api.Rule, tlsConfig
 	rules = AddDefaultRouteRules(rules, instances)
 
 	backends := make(map[string]*api.Backend)
+	staticClusters := make(map[string]*api.Backend)
```
ijsnellf: The logic from line 336 to 409 could be simplified by making `staticClusters` a set.

sidecar/proxy/envoy/manager.go
```diff
@@ -403,15 +408,84 @@ func BuildWeightKey(service string, tags []string) string {
...
+
+	return cluster
+}
+
 // BuildRoutes builds routes based on rules.  Assumes at least one route for each service
 func BuildRoutes(ruleList []api.Rule, instances []*api.ServiceInstance) []Route {
 	SanitizeRules(ruleList)
 	rules := AddDefaultRouteRules(ruleList, instances)
 
+	staticClusters := make(map[string]*api.Backend)
```
ijsnellf: We never use the map value `*api.Backend`, so `staticClusters` would probably be more appropriate as a set: `map[string]struct{}`

ijsnellf (changes requested):
Could be simplified a bit.

sidecar/proxy/envoy/manager.go
```diff
@@ -330,65 +333,67 @@ func BuildClusters(instances []*api.ServiceInstance, rules []api.Rule, tlsConfig
...
-			}
+		clusters = append(clusters, cluster)
+	}
 
-			// Envoy outlier detection settings that complete circuit breaker
-			if backend.Resilience.SleepWindow > 0 {
-				cluster.OutlierDetection.BaseEjectionTimeMS = int(backend.Resilience.SleepWindow * 1000)
+	for clusterName, backend := range staticClusters {
+		cluster := buildCluster(clusterName, backend, tlsConfig)
+		cluster.Type = "strict_dns"
```
ijsnellf: `buildStaticCluster` function?

ijsnellf (approved):

PR #510, opened 2017-02-09
Enable helloworld testcase for examples automation.

craigyam:
The helloworld tests are only run for examples automation
(make test.examples).  Re-enabling them to run for examples
using the k8s native routing rules.  The integration tests for
k8s will still be disabled until the bookinfo tests are updated.

### DIFF
--- a/testing/test-scripts/helloworld-v1-v2-route-rules.yaml
+++ b/testing/test-scripts/helloworld-v1-v2-route-rules.yaml
@@ -24,5 +24,3 @@ spec:
       - weight: 0.25
         tags:
         - "version=v2"
-      - tags:
-        - "version=v1"
--- a/testing/test-scripts/helloworld-v1-v2-route-rules.yaml
+++ b/testing/test-scripts/helloworld-v1-v2-route-rules.yaml
@@ -24,3 +24,5 @@ spec:
       - weight: 0.25
         tags:
         - "version=v2"
+      - tags:
+        - "version=v1"
--- a/testing/test-scripts/helloworld.sh
+++ b/testing/test-scripts/helloworld.sh
@@ -14,16 +14,17 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-set -o errexit
-
 # docker or k8s
-ENV=$1
+if [ -z "$A8_CONTAINER_ENV" ]; then
+    A8_CONTAINER_ENV="docker"
+fi
+ENV=$A8_CONTAINER_ENV
 
 # Dump some possibly useful data on failures to help debug
 dump_debug() {
-    curl localhost:31200/v1/rules
 
     if [ "$ENV" == "docker" ]; then
+        curl localhost:31200/v1/rules
         docker exec -it gateway a8sidecar --debug show-state
         docker logs gateway
         docker exec gateway ps aux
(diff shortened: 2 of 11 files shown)


testing/test-scripts/helloworld-v1-v2-route-rules.yaml
```diff
@@ -24,3 +24,5 @@ spec:
       - weight: 0.25
         tags:
         - "version=v2"
+      - tags:
+        - "version=v1"
```
elevran: Don't understand why additional tag needed here.
Wanting to have v2 with 25% probability. Needs priority fix as in #512.

testing/test-scripts/helloworld.sh
```diff
@@ -51,26 +53,32 @@ MAX_LOOP=5
...
+    cleanup_all_rules
 
     echo ""
     echo "Set/verify the Hello World default route to v1"
-    echo "  Invoking a8ctl to set the default route"
-    $A8CLI rule-create -f $SCRIPTDIR/helloworld-default-route-rules.json > /tmp/a8ctl_output.txt
+    if [ "$ENV" == "docker" ]; then
+        create_rule $SCRIPTDIR/helloworld-default-route-rules.json
+    else
+        create_rule $SCRIPTDIR/helloworld-default-route-rules.yaml
```
elevran: At least currently we can create the json and yaml rules from the same source more or less. Something to consider so we don't have duplicate code?

testing/test-scripts/helloworld.sh
```diff
@@ -82,9 +90,6 @@ while [  $retry_count -le $((MAX_LOOP)) ]; do
         echo ""
     fi
 
-    # Sleep a bit to make sure the rules are applied
```
elevran, line 85: Envoy still takes sometime to restart. Don't we need the sleep?

testing/test-scripts/helloworld.sh
elevran: Also, seems like controlplane and demo topologies are copied to examples using a `docker` or `k8s` prefix, but rules are copied as is. In case there is ever a divergence, we should consider adding a prefix to the rules on copying as well.

elevran (commented):
Overall looks ok. Some clarifications needed per comments.

testing/test-scripts/helloworld-v1-v2-route-rules.yaml
craigyam: The script clears out the previous rules before the next tests.  Without setting version=v1 in that case 3/4 of the requests get a 404.  I can update the script to keep the default routing rule and then remove that.

testing/test-scripts/helloworld.sh
craigyam: The rules are currently not copied over to examples/ at all.  I'll update the automation to do that with the k8s prefix.

testing/test-scripts/helloworld-v1-v2-route-rules.yaml
elevran: Or add the previous rule as well. Your call.

elevran (approved):
LGTM once comments are integrated into this PR

testing/test-scripts/helloworld-v1-v2-route-rules.yaml
craigyam: Done

testing/test-scripts/helloworld.sh
craigyam: Will investigate generating rules files for automation.  For now, copy to examples/ and prefix appropriately.

testing/test-scripts/helloworld.sh
craigyam: Sleep was added at 76 after rules creation.

PR #512, opened 2017-02-09
set example rule priority, allow priority to stay unspecified

elevran:
Set v2-v1 rule priority higher so it has precedence over default v1 routes
Remove priority as required rule attribute to match documentation


zcahana (approved):
LGTM.
It may also be good time to update the schema doc on the website:
https://github.com/amalgam8/amalgam8.github.io/blob/master/_api/controller-rules-schema.md
https://www.amalgam8.io/api/controller-rules-schema/

PR #520, opened 2017-02-19
Use k8s replicasets

elevran:
Replace ReplicationController with ReplicaSet


testing/kubernetes/bookinfo.yaml
```diff
@@ -31,7 +31,7 @@ spec:
 ---
 
 apiVersion: v1
-kind: ReplicationController
+kind: ReplicaSet
```
zcahana: The `ReplicaSet` kind is still an extension API object, so you have to use `apiVersion: extensions/v1beta1`. See [here](https://kubernetes.io/docs/user-guide/replicasets/).

zcahana (changes requested):

testing/kubernetes/bookinfo.yaml
elevran: Correct. Fixed and resubmitted

zcahana (approved):

zcahana:
LGTM. However, note that the usage of `selector:` here, while not incorrect, is redundant, as it defaults to `matchLabels` on the template pod labels.

PR #523, opened 2017-02-21
bug fix: subset matching for tags if instances has more than one

GregHanson:
kubernetes adapter appends additional tags to service instance.

Example:

```
a8ctl-beta-linux traffic-start -s reviews -v version=v3
```

Should the traffic-start command only match services registered as `reviews:version=v3` or should it also match `reviews:version=v3,tag2=value2` ?


gicaceres (approved):
LGTM

PR #528, opened 2017-02-24
update envoy binary

rshriram:


rshriram:
This version has support for routing to cloud foundry instances.

You can specify hostnames in cluster, with strict_dns option.. and then in the route block, use auto_host_rewrite. When auto_host_rewrite is set to true, envoy will automatically rewrite the host header with the name of the upstream host (in this case the cloud foundry app name).

```diff
+                    },
+                    {
+                      "timeout_ms": 0,
+                      "prefix": "/rewritebin",
+                     "prefix_rewrite" : "/",
+                     "auto_host_rewrite" : true,
+                      "cluster": "freebin"
                     }
 
                   ]
@@ -71,6 +78,17 @@
             "url": "tcp://service2:80"
           }
         ]
+      },
+      {
+        "name": "freebin",
+        "connect_timeout_ms": 250,
+        "type": "strict_dns",
+        "lb_type": "round_robin",
+        "hosts": [
+          {
+            "url": "tcp://mycfapp.mybluemix.net:80"
+          }
+        ]
       }
```

For example, in the above diff, the top half shows a route configuration. calls to localhost:6379/rewritebin get routed to mycfapp.mybluemix.net .. While doing so, envoy rewrites the host header from `Host: localhost:6379` to `Host: mycfapp.mybluemix.net` .. Without the proper Host header, requests to CF apps won't work as they will be dropped on the floor by the GoRouter.

cc: @iwinoto @roder ^^^

craigyam (approved):

PR #534, opened 2017-03-03
k8s controller backend

ijsnellf:
1st pass at adding support for a kubernetes backed controller. The controller does not support bulk operations. Additionally, this PR does not modify a8ctl to communicate with the kubernetes registry.


ijsnellf:
@elevran: I accidentally wiped your comments with a `push -f` while I was refactoring the commit. My bad. :-/ I believe I addressed all your comments except for the suggestion to use a watcher/cache to speed up retrieval times.

controller/config/config.go
```diff
@@ -25,13 +25,20 @@ import (
 	"github.com/urfave/cli"
 )
 
+// Database types
+const (
+	DatabaseTypeRedis  = "redis"
+	DatabaseTypeMemory = "memory"
+	DatabaseTypeK8S    = "k8s"
```
elevran: If this is exposed to the user, I think we're using  `kubernetes` elsewhere in config/flag values, so consider changing for consistency. Should be reflected in docs too.

controller/rules/k8s.go
```diff
@@ -0,0 +1,182 @@
...
+
+var (
+	// BulkNotSupportedError is returned when unsupported bulk operations are attempted.
+	BulkNotSupportedError = fmt.Errorf("bulk operations not supported")
+)
+
+// K8S controller. Currently no bulk insert/update operations are supported and
+// the controller is bound to a single namespace.
+type K8S struct {
+	// kubernetes REST client
```
elevran: Doc comment not strictly required (since not exported). Not sure what "the style guide" says about whitespace... Ditto below

controller/rules/k8s.go
```diff
@@ -0,0 +1,182 @@
...
+
+	return &K8S{
+		client:    client,
+		validator: validator,
+		namespace: ns,
+	}, nil
+}
+
+// AddRules validates the rules and adds them to the collection for the namespace.
+// TODO: bulk operations
```
elevran: Is it really a TODO? Are we keeping bulk updates around or removing from all?

testing/test-scripts/bookinfo-default-route-details.yaml
```diff
@@ -0,0 +1,7 @@
+rules:
+- priority: 1
+  destination: details
+  route:
+    backends:
+    - tags:
+      - version=v1
```
elevran: seems to indicate missing newline at end of file

testing/test-scripts/testing-common.sh
```diff
@@ -15,65 +15,24 @@
 #   limitations under the License.
 
 cleanup_all_rules() {
-    if [ "$ENV" == "docker" ]; then
-        echo "Cleaning up docker rules"
-        $A8CLI rule-delete -a -f
+    echo "Cleaning up docker rules"
```
elevran: `docker rules` or just `rules`?

elevran (approved):

PR #535, opened 2017-03-06
Removed traffic commands

ijsnellf:
Removed traffic related commands: `traffic-set`, `traffic-step` etc


rshriram (approved):
Why remove the checks in the tests? At the very least let's comment it out. I thought we could do a create rule and check it with the existing code in the test scripts. @craigyam ?

ijsnellf:
@rshriram: I replaced the existing tests which were mainly focused on the traffic commands with a single test that tests percentage based routing. As for commenting out vs deleting --we have source control. If we ever decide to add them back, we can just look at the git history for before this PR was merged.

GregHanson:
@rshriram it is also not easy to implement `traffic-step` with `rule-create` and `rule-update` commands.  Updating the existing rule would require us to track the rule ID (in bash nonetheless) during the the subsequent steps.  Creating a new rule every time would require a `rule-delete -a -f` and then we would also have to re-create the default routing rules during every step of the process

GregHanson (approved):
LGTM

ijsnellf:
We could replicate all the `traffic-step` etc commands with `rule-create` and `rule-update` cmds but I don't see what we gain. They simply test if 10%, then 20%, and so on of the traffic is routed. The same cases are covered by one 25/75 test.

PR #536, opened 2017-03-06
Removed registry logic from a8ctl

ijsnellf:


rshriram (approved):

cli/README.md
```diff
@@ -19,16 +19,12 @@ $ a8ctl-beta -h
 
 Commands
 --------
-- a8ctl-beta service-list
 - a8ctl-beta route-list
```
ijsnellf, line 22: Perhaps as a separate PR we should nuke these readme files because they are out of date and not used anywhere anymore.

mossuchida:
I still see service-list in https://www.amalgam8.io/docs/demo-helloworld.html.  Can you also update this html page also?

PR #537, opened 2017-03-06
Remove the "set" API

ijsnellf:
The "set" API is not worth maintaining and cannot be implemented for some backends.


GregHanson (approved):
LGTM

PR #539, opened 2017-03-07
configurable rule ids

GregHanson:


controller/rules/redismanager.go
```diff
@@ -51,14 +51,16 @@ func (r *redisManager) AddRules(namespace string, rules []api.Rule) (NewRules, e
 
 	entries := make(map[string]string)
 	for i := range rules {
-		id := uuid.New() // Generate an ID for each rule
-		rules[i].ID = id
+		if rules[i].ID == "" {
+			id := uuid.New() // Generate an ID for each rule
+			rules[i].ID = id
```
ijsnellf: No need for extra variable here:
`rules[i].ID = uuid.New()`

ijsnellf (approved):

rshriram (approved):
LGTM.. Don't you have to check if the rule exists or not? Secondly, what if we mandate the ID from end user instead of us generating the ID?

PR #545, opened 2017-03-07
Updated changelog

ijsnellf:
Please look this over.


rshriram (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+## 1.0.0 (March 8, 2017)
```
elevran, line 3: are we calling it 1.0? why not 0.5.0 as a follow on to 0.4.x?

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+
+- Support for gRPC over HTTP/1.1
+
+- Support for configuring circuit breakers, bulkheads, timeouts, retries
+
+- Support for different load balancing algorithms (round robin, least conn, random) ([PR#494](https://github.com/amalgam8/amalgam8/pull/494))
+
+- Deeper integration with Kubernetes: rules are now stored as third party resources ([PR#534](https://github.com/amalgam8/amalgam8/pull/534))
+
+- A fully revamped go based cross platform CLI (a8ctl) ([PR#396](https://github.com/amalgam8/amalgam8/pull/396))
```
elevran: Capitalize Go

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+## 1.0.0 (March 8, 2017)
+
+- Switched amalgam8 sidecar to use Envoy proxy from Lyft (https://github.com/lyft/envoy)
+
+- Support for HTTP/2, gRPC proxying (downstream and upstream)
+
+- Support for gRPC over HTTP/1.1
```
elevran: IIRC. Envoy also supports protocol bridging between gRPC on 1.1 and 2.0. Please confirm with @rshriram

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+## 1.0.0 (March 8, 2017)
+
+- Switched amalgam8 sidecar to use Envoy proxy from Lyft (https://github.com/lyft/envoy)
```
elevran: I'm for `amalgam8` but the docs seem to capitalize it as `Amalgam8`. Should be consistent (but I would NOT go over all docs and change the letter case...)

CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+
+- Support for HTTP/2, gRPC proxying (downstream and upstream)
+
+- Support for gRPC over HTTP/1.1
+
+- Support for configuring circuit breakers, bulkheads, timeouts, retries
+
+- Support for different load balancing algorithms (round robin, least conn, random) ([PR#494](https://github.com/amalgam8/amalgam8/pull/494))
+
+- Deeper integration with Kubernetes: rules are now stored as third party resources ([PR#534](https://github.com/amalgam8/amalgam8/pull/534))
```
elevran: Consider adding that can use kubectl to create rules directly and link to docs. We have two modes for k8s: with and without the rules controller. With is the preferred mode, so on second thought, maybe don't mention direct access using kubectl.

elevran (approved):
some nits, otherwise LGTM and can merge

CHANGELOG.md
GregHanson: @elevran following this standard:
```
Given a version number MAJOR.MINOR.PATCH, increment the:

    MAJOR version when you make incompatible API changes,
    MINOR version when you add functionality in a backwards-compatible manner, and
    PATCH version when you make backwards-compatible bug fixes.
```
I would not be opposed to 1.0 over 1.0.0

ijsnellf: The main argument against 1.0.0 as per semantic versioning is that some of our previous releases haven't followed that standard. 0.1.x to 0.2.0 had breaking changes, for instance.

ijsnellf (commented):
Nit: in the rest of our changelogs, we put a period at the end of all the bullets.

PR #546, opened 2017-03-08
Fix a few issues with the bookinfo automation script.

craigyam:
- Failing weighted routing test does not fail overall test
- Failing weighted routing tests do not retry correctly
- Additional rules were causing the routing to not work correctly.


rshriram (commented):
Now do the weighted routing rules work properly?

elevran (approved):
LGTM

PR #547, opened 2017-03-08
updated envoy binary with outlier detection and circuit breaker fixes

GregHanson:
included a more verbose rule for productpage with all resilience fields set to ensure envoy accepts them in their schema

### DIFF
--- a/testing/test-scripts/bookinfo-default-route-productpage.yaml
+++ b/testing/test-scripts/bookinfo-default-route-productpage.yaml
@@ -6,3 +6,11 @@ rules:
     backends:
     - tags:
       - version=v1
+      resilience:
+        max_connections: 1111
+        max_pending_requests: 222
+        max_requests: 222
+        sleep_window: 0.9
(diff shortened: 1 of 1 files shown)


testing/test-scripts/bookinfo-default-route-productpage.yaml
```diff
@@ -6,3 +6,11 @@ rules:
     backends:
     - tags:
       - version=v1
+      resilience:
```
ijsnellf, line 9: Even though these aren't things that are directly triggered, these files do get used by the demos, so why don't we have less strange values? Otherwise people will look at them and be confused about the significance of 1111 vs simply 1000.

1111 -> 1000
222 -> 200
1234 -> 1500
etc

ijsnellf:
Even though these aren't things that are directly triggered, these files do get used by the demos, so why don't we have less strange values?

ijsnellf (approved):

Issue #459, opened 2017-01-05
Add TCP support to Envoy proxy adapter.


ijsnellf:
Envoy supports TCP. The Envoy proxy adapter should be modified to leverage Envoy for TCP endpoints.

jemcginn:
P3 - Not needed for bookinfo or Nginx parity

PR #548, opened 2017-03-10
Add TCP support to Envoy proxy adapter

kimikowang:
Issue #459


sidecar/config/config.go
```diff
@@ -167,6 +172,8 @@ type Config struct {
 
 	ProxyConfig ProxyConfig `yaml:"proxy_config"`
 
+	TCPProxyConfigs []TCPProxyConfig `yaml:"tcp_proxy_configs"`
+
```
GregHanson: The `TCPProxyConfigs` should probably be added under the `ProxyConfig` object since it is only used in the proxy code.  And to avoid some repetition in the name it should probably be called just `TCPProxies`

sidecar/proxy/envoy/resources.go
```diff
@@ -108,12 +108,26 @@ type VirtualHost struct {
...
+	DestinationIPList []string `json:"destination_ip_list,omitempty"`
+	DestinationPorts  string   `json:"destination_ports,omitempty"`
+	SourceIPList      []string `json:"source_ip_list,omitempty"`
+	SourcePorts       string   `json:"source_ports,omitempty"`
+}
+
+// HttpRouteConfig definition.
 // See: https://lyft.github.io/envoy/docs/configuration/http_conn_man/route_config/route_config.html#config-http-conn-man-route-table
-type RouteConfig struct {
+type HttpRouteConfig struct {
```
GregHanson: name should be HTTPRouteConfig to satisfy `make precommit` conditions

sidecar/proxy/envoy/resources.go
```diff
@@ -129,24 +143,30 @@ type RDS struct {
...
-	CodecType         string       `json:"codec_type"`
-	StatPrefix        string       `json:"stat_prefix"`
-	GenerateRequestID bool         `json:"generate_request_id"`
-	UserAgent         bool         `json:"add_user_agent"`
-	RouteConfig       *RouteConfig `json:"route_config,omitempty"`
-	RDS               *RDS         `json:"rds,omitempty"`
-	Filters           []Filter     `json:"filters"`
-	AccessLog         []AccessLog  `json:"access_log"`
+// HttpFilterConfig definition.
+type HttpFilterConfig struct {
```
GregHanson: HTTPFilterConfig

testing/apps/tcp/app.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
+#
+# import socket
+#
+# TCP_IP = '127.0.0.1'
+# TCP_PORT = 1111
+# BUFFER_SIZE = 20  # Normally 1024, but we want fast response
+#
+# s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
GregHanson: remove commented code

testing/apps/tcp/test.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
+#
+# import socket
+#
+#
+# TCP_IP = '127.0.0.1'
+# TCP_PORT = 1111
+# BUFFER_SIZE = 1024
```
GregHanson: remove commented code

testing/apps/tcp/details/details.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2016 IBM Corporation
```
GregHanson: is this tcp version of details used in testing at all?

GregHanson (changes requested):
I would recommend adding at least one docker test case to the Travis CI test scenario to make sure our TCP envoy solution works.  A test case for Kubernetes can be added in this PR or as a separate pull request

sidecar/proxy/envoy/manager.go
```diff
@@ -283,6 +246,90 @@ func (m *manager) generateConfig(rules []api.Rule, instances []api.ServiceInstan
 	}, nil
 }
 
+func BuildRDSClusters(tcpProxyList []config.TCPProxyConfig, sdsPort int) []Cluster {
```
GregHanson: function should be named `buildStaticClusters` since it is building more than just the `rds` cluster and it shouldn't need to be exported since it is only called in `generateConfig()`

GregHanson (approved):

testing/docker/bookinfo.yaml
```diff
@@ -33,6 +33,19 @@ gateway:
   container_name: gateway
   command: /usr/bin/run_filebeat.sh
 
+tcphelloworld:
+  image: amalgam8/a8-tcp:latest
+  environment:
+    - A8_REGISTRY_URL=http://registry:8080
+    - A8_SERVICE=tcphelloworld
```
rshriram: Please attach a tag to this service name.
e.g.
`A8_SERVICE=tcphelloworld:version=v1`

testing/apps/productpage/config.yaml
```diff
@@ -4,3 +4,8 @@ commands:
     on_exit: ignore
   - cmd: [ "/opt/microservices/productpage", "9080", "http://localhost:6379" ]
     on_exit: terminate
+
+proxy_config:
+  tcp_proxies:
+    - service: tcphelloworld
```
rshriram: change to use tags, e.g., `tcphelloworld:version=v1`

testing/apps/productpage/Dockerfile.sidecar
```diff
@@ -15,8 +15,8 @@
 FROM amalgam8/a8-sidecar:latest
 
 WORKDIR /opt/microservices
+COPY config.yaml /opt/microservices/
 COPY productpage /opt/microservices/
-COPY config.yaml /opt/microservices/config.yaml
-EXPOSE 9080
+EXPOSE 9080 12345
```
rshriram, line 20: why should productpage expose 12345 ? Its only for productpage app to Envoy. Nothing is going to talk to productpage over 12345

testing/apps/tcp/Makefile
```diff
@@ -0,0 +1,18 @@
+APP_NAME 	:= $(shell basename `pwd`)
+IMAGE_NAME 	:= amalgam8/a8-$(APP_NAME)
```
rshriram: If both client and server are in python, why do we need go based build in teh makefile?
All you need is the dockerfiles and a makefile to create those. i.e. you don't need build targets or build flags or any go related stuff.

testing/docker/bookinfo.yaml
```diff
@@ -115,6 +128,8 @@ productpage-v1:
     - A8_REGISTRY_POLL=5s
     - A8_HEALTHCHECKS=http://localhost:9080/health,file:///bin/ls
     - A8_ELASTICSEARCH_SERVER=elasticsearch:9200
+  ports:
+    - "12345:12345"
```
rshriram, line 132: why is this needed?

sidecar/proxy/envoy/manager.go
```diff
@@ -196,59 +198,47 @@ func (m *manager) generateConfig(rules []api.Rule, instances []api.ServiceInstan
...
+				},
+			},
+			MaxRequestsPerConnection: 1,
+		},
+	}
+
+	for _, proxy := range m.tcpProxyConfigs {
+		cluster := Cluster{
+			Name:             proxy.Service,
+			ServiceName:      proxy.Service,
```
rshriram, line 223: Why do you need to build clusters for TCP? CDS is already going to be pulling in all the clusters from the service registry. All you need is to build listeners.

rshriram: The reason I ask is because we have circuit breaker settings. And we need to be able to configure circuit breakers for these services.

rshriram (changes requested):

sidecar/proxy/envoy/manager.go
GregHanson: TCP filters have a statically defined cluster name.  Envoy rejects the config if there is not a statically defined cluster in cluster_manager.  CDS is not compatible with TCP at this point

testing/apps/productpage/Dockerfile.sidecar
GregHanson: 12345 is a hack to test that the envoy proxy inside productpage can reach the tcphelloworld app in tests.  tcp listeners need to be configured via a config.yaml in sidecar.  We didn't want to modify the gateway image, so we modified productpage since it already had a config yaml to accomplish the same thing

rshriram (approved):

PR #549, opened 2017-03-17
first pass at fixing tls validation logic and where ssl context is added

GregHanson:


rshriram (approved):
@GregHanson merge when you see fit. And please update Envoy image.

PR #553, opened 2017-03-27
Tar the cli binaries with a8ctl name.

craigyam:
Name the binary a8ctl but tar/zip with the platform.


craigyam:
We'll just update the docs instead to tell the user to rename the file.

